PR #1, opened 2024-11-05
RSDK-9160 lorawan OTAA join procedure

oliviamiller:
The gateway module parses a join request message from a device, then constructs and sends a join accept message. The app session key is calculated and used to decrypt data uplinks from the device. The bytes sent in the data uplink are decoded using the decoder provided in the config.

Note that the config and models don't match the scope yet, this PR is just for getting basic OTAA join procedure flow completely working. 

Tested on milesight ct101 and dragino temp sensor, ABP and OTAA both work!


gateway/join.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	var cPayload [256]C.uchar
+	for i, b := range joinAccept {
+		cPayload[i] = C.uchar(b)
+	}
+	txPkt.payload = cPayload
+
+	// send on rx2 window - opens 6 seconds after join request.
+	time.Sleep(joinRx2WindowSec * time.Second)
```
JohnN193, line 75: SelectContrxtOrWait or similar so we can cancel the sleep

gateway/join.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// match the dev eui to gateway device
+	for _, device := range devices {
+		if bytes.Equal(device.devEui, devEUIBE) {
+			matched = device
+		}
+	}
+
+	if matched.name == "" {
+		return JoinRequest{}, nil, errors.New("received join request from unknown device")
```
JohnN193, line 113: Can we return the name given in the request?

gateway/sensor.go
```diff
@@ -174,139 +232,85 @@ func (g *Gateway) receivePackets() {
...
-		if err != nil {
-			g.logger.Errorf("error parsing uplink message: %w", err)
-		} else if name != "" {
-			// the milesight lora sensor sends two data uplink messages close together, we want both of these in the map.
+func (g *Gateway) handlePacket(ctx context.Context, payload []byte) {
+	g.workers.Add(func(ctx context.Context) {
+		// first byte is MHDR - specifies message type
+		switch payload[0] {
+		case 0x0:
+			g.logger.Infof("recieved join request")
```
JohnN193, line 268: Should these be debug logs? Also fine if you wanna leave while the project is in progress

JohnN193 (commented):
Took a first glance with some thoughts, will look again tomorrow

gateway/sensor.go
oliviamiller: yeah I can change them to debug later on

gateway/join.go
JohnN193: So there’s no info about the device that’s sending data to us? 

Is it possible we would hit this error if a user misconfigured their devices in Viam?

oliviamiller: The identifying info in the join request is the dev EUI. If the user put in the wrong EUI into the config they would hit this. They would also hit this if there is another lorawan device nearby that the user doesn't care about sending messages.

gateway/uplink.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/robertkrimen/otto"
+	"go.thethings.network/lorawan-stack/v3/pkg/crypto"
+	"go.thethings.network/lorawan-stack/v3/pkg/types"
+)
+
+// Structure of phyPayload:
+// | MHDR | DEV ADDR|  FCTL |   FCnt  | FPort   |  FOpts     |  FRM Payload | MIC |
+// | 1 B  |   4 B    | 1 B   |  2 B   |   1 B   | variable    |  variable   | 4B  |
+func (g *Gateway) parseDataUplink(phyPayload []byte) (string, map[string]interface{}, error) {
```
randhid, line 19: are your helpers easy to set up tests for at this point?

gateway/join.go
```diff
@@ -0,0 +1,254 @@
...
+	rand := rand.New(source)
+
+	num1 := rand.Intn(255)
+	num2 := rand.Intn(255)
+	num3 := rand.Intn(255)
+
+	return []byte{byte(num1), byte(num2), byte(num3)}
+}
+
+// reverseByteArray creates a new array reversed of the input.
```
randhid, line 246: What/why do we need a reversed byte array for add to comment

gateway/join.go
```diff
@@ -0,0 +1,254 @@
...
+	joinRx2WindowSec = 6         // rx2 delay for sending join accept message.
+	rx2Frequenecy    = 923300000 // Frequuency for rx2 window
+	rx2SF            = 12        // spreading factor for rx2 window
+	rx2Bandwidth     = 0x06      // 500k bandwidth
+)
+
+var errNoDevice = errors.New("received join request from unknown device")
+
+// network id for the device to identify the network.
+var netID = []byte{1, 2, 3}
```
randhid, line 45: why 1,2,3? Is it significant for the order? If we need a byte array that's just empty by 3 elements 3, could we do var netID [3]byte?

gateway/uplink.go
```diff
@@ -0,0 +1,148 @@
...
+		return map[string]interface{}{}, errors.New("decoder returned unexpected data type")
+	}
+
+	readings := v.(map[string]interface{})
+
+	return readings, nil
+}
+
+func executeDecoder(script string, vars map[string]interface{}) (out interface{}, err error) {
+	defer func() {
```
randhid, line 117: Does this function benefit from a context so that the go function is interruptable by a cancel? Similar to Generate Join Accept.

gateway/join.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	return joinRequest, matched, nil
+
+}
+
+// Format of Join Accept message:
+// | MHDR | JOIN NONCE | NETID |   DEV ADDR  | DL | RX DELAY |   CFLIST   | MIC  |
+// | 1 B  |     3 B    |   3 B |     4 B     | 1B |    1B    |  0 or 16   | 4 B  |
+func (d *Device) GenerateJoinAccept(ctx context.Context, jr JoinRequest) ([]byte, error) {
```
randhid, line 133: Any reason this is the only exported function in this diff?

gateway/gateway.c
```diff
@@ -1,30 +1,110 @@
...
+    boardconf.clksrc = 0;
+    boardconf.full_duplex = false;
+    boardconf.com_type =  LGW_COM_SPI; // spi
+
+    const char * com_path;
+    if (bus == 1) {
+        com_path = "/dev/spidev0.1";
+    } else {
+        com_path = "/dev/spidev0.0";
+    }
```
randhid, line 50: convert to a switch statement with a default and a bus ==1 case.

randhid (commented):
Overall super clean. I'd like more comments throughout explaining the flow and if anything is in a state where you want to prevent regressions in logic, add tests.

gateway/join.go
oliviamiller: its just a random number I put, we just need a 3 byte ID to send the devices that's used to identify our network.

gateway/join.go
oliviamiller: no ill unexport

gateway/gateway.c
```diff
@@ -1,30 +1,110 @@
...
+    rfconf.radio_type = LGW_RADIO_TYPE_SX1250;
+    rfconf.rssi_offset = -215;
+    rfconf.tx_enable = true;
+
+    if (lgw_rxrf_setconf(0, &rfconf) != LGW_HAL_SUCCESS) {
+        return EXIT_FAILURE;
+    }
+
+    rfconf.freq_hz = RADIO_1_FREQ;
+    if (lgw_rxrf_setconf(1, &rfconf) != LGW_HAL_SUCCESS) {
```
martha-johnston, line 71: why do both freq 0 and 1 need to succeed? maybe add a comment?

gateway/gateway.c
```diff
@@ -1,30 +1,110 @@
...
+};
+
+// This defines what RF chain to use for each of the 8 if chains
+const int32_t rfChains [9] = {0, 0, 0, 0, 1, 1, 1};
+
+
+int setUpGateway(int bus) {
+    struct lgw_conf_board_s boardconf;
+    struct lgw_conf_rxrf_s rfconf;
+    struct lgw_conf_rxif_s ifconf;
```
martha-johnston, line 37: what's the difference between rfconf and ifconf? I'm sure this (and many of my questions) will be because I'm not familiar with the hardware, so maybe add a link to helpful documentation if it exists?

gateway/gateway.c
```diff
@@ -1,30 +1,110 @@
...
+
+    // set config for intermediate frequency chains.
+    memset(&ifconf, 0, sizeof(ifconf));
+    ifconf.enable = true;
+    ifconf.datarate = DR_LORA_SF7;
+    ifconf.bandwidth = 0x04; //125k
+    for (int i = 0; i < 8; i++) {
+        ifconf.rf_chain = rfChains[i];
+        ifconf.freq_hz = ifFrequencies[i];
+        if (lgw_rxif_setconf(i, &ifconf) != LGW_HAL_SUCCESS) {
```
martha-johnston, line 84: same question here about why it has to succeed on all frequencies. I'm assuming one is sending info and one is receiving, and you want to make sure it's able to do its job on all available frequencies? if I'm wrong consider adding a comment about what's happening here?

gateway/join.go
```diff
@@ -0,0 +1,254 @@
...
+	if !utils.SelectContextOrWait(ctx, time.Second*joinRx2WindowSec) {
+		return nil
+	}
+
+	// lock so there is not two sends at the same time.
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	errCode := int(C.send(&txPkt))
+	if errCode != 0 {
+		return errors.New("failed to send join accept packet")
```
martha-johnston, line 87: for this and all other errors, consider adding at the top as vars. or remove `errNoDevice` and create the error when it occurs for consistency

gateway/gateway.c
oliviamiller: I will add a lot more comments to this file to explain each step, basically there are two RF (radio frequency) chains and 8 IF (intermediate frequency) chains on the gateway that all need to be configured. This is what allows the gateway to be able to receive and transmit at multiple different frequencies.

gateway/sensor.go
```diff
@@ -68,22 +67,88 @@ func init() {
...
+				errors.New("decoder path is required"))
+		}
+		switch d.JoinType {
+		case "ABP":
+			return d.validateABPAttributes(path)
+		case "OTAA", "":
+			return d.validateOTAAAttributes(path)
+		default:
+			return nil, resource.NewConfigValidationError(path,
+				errors.New("join type is OTAA or ABP"))
```
martha-johnston, line 82:
```suggestion
				errors.New("join type must be OTAA or ABP"))
```

martha-johnston: consider adding the note about the default, might be helpful for someone who hits this error and isn't sure where to go from there

martha-johnston (approved):
agree with adding more comments, but otherwise lgtm!

gateway/join.go
randhid: So i could be a [3]byte instantiated in go, but it is totally okay to keep it as is with a callout that it has to be a length 3 array, and that the number values don't mean anything for instantiation.

gateway/uplink.go
oliviamiller: I will write tests after the split of device/gateway to avoid having to change the structure of the tests

gateway/join.go
oliviamiller: I created a variable for errNoDevice because we check for it and ignore that error in `handlePacket`. Don't think its worth doing for others that we won't need to use again.

gateway/join.go
JohnN193: so is it worth returning the request's EUI as a part of the error, to help the user identify if their device is misconfigured or if theres another device interacting with the network that they are not aware of? 

I do not know how common it is to have outside devices pinging a network on the same frequencies, but I figure if a user misconfigured a device then returning the EUI should at least speed up debugging.

if you don't think its worthwhile feel free to ignore

JohnN193 (approved):

gateway/join.go
oliviamiller: It could be common if they are in an area with a lot of other lora sensors, I will add a debug log with the dev EUI.

gateway/join.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// everything but the mhdr needs to be encrypted.
+	payload = payload[1:]
+
+	payload = append(payload, resMIC[:]...)
+
+	enc, err := crypto.EncryptJoinAccept(d.AppKey, payload)
```
randhid, line 167: Do we want to completely swallow the error? I'd log it if so, even if Debug, or use errors.Join to append multiple errors together if we should continue even in an error state.

gateway/uplink.go
```diff
@@ -0,0 +1,163 @@
...
+	return device.name, readings, nil
+}
+
+func matchDeviceAddr(devAddr []byte, devices map[string]*Device) (*Device, error) {
+	for _, dev := range devices {
+		if bytes.Equal(devAddr, dev.addr) {
+			return dev, nil
+		}
+	}
+	return nil, errors.New("no match")
```
randhid, line 76:
```suggestion
	return nil, fmt.errors("no match for DeviceAddress %v", devAddr)
```

gateway/sensor.go
```diff
@@ -68,22 +67,88 @@ func init() {
...
 	resource.Named
 	resource.AlwaysRebuild
 	logger logging.Logger
 
 	workers *utils.StoppableWorkers
+	mu      sync.Mutex
 
-	lastReadings map[string]interface{} // map devices to their last reading
+	lastReadings map[string]interface{} // map of devices to readings
+	readingsMu   sync.Mutex
```
randhid, line 148: keep an eye on the two mutexes as you code more; if it makes sense to move the readings mutex and last reading into a struct that has a separate mutex hold to write and overwrite, we should do that early in this module's life.

Places we care about double locking the struct are the constructor + `Reconfigure`, `Close`, and `Readings` being spammed by multiple clients + data capture.

randhid (approved):
Honestly LGTM, one nit on the line you're swallowing an error, and a question on why the double mutex in the sensor struct.

gateway/join.go
oliviamiller: oops no that was not intentional, we should return the error here. fixing

gateway/sensor.go
oliviamiller: the `mu` mutex is to avoid sending on a gateway channel at the same, but I think that we should be able to get the readings at the same time as a send hence the two mutexes. But maybe just one would be sufficient.

PR #2, opened 2024-11-11
RSDK-9207 RSDK-9208 Make separate node model, add gateway config attributes 

oliviamiller:
Seperate gateway and node into two models. 

Node model sends a `validate` docommand to its dependency to validate its the gateway. It then sends `register` docommand to register itself as a known component with the gateway. Readings calls the gateway readings and gets its value from the map.

Added config attributes for spi bus and pins to gateway model.


node/node.go
```diff
@@ -0,0 +1,242 @@
...
+	if n.gateway != nil {
+		allReadings, err := n.gateway.Readings(ctx, nil)
+		if err != nil {
+			return map[string]interface{}{}, err
+		}
+
+		reading, ok := allReadings[n.NodeName]
+		if !ok {
+			// no readings availiable yet
+			return map[string]interface{}{}, data.ErrNoCaptureToStore
```
oliviamiller, line 236: This will happen if the lorawan sensor hasn't sent any data yet. Returning this error will prevent an error from spamming in the logs if data capture is on or you're on the control page. However, the error is `no capture from filter module` which does not reflect what's actually happening and may confuse users. Would like reviewers thoughts on what the behavior should be here.

randhid: If we add a log line before it, does that clarify things, can we join this error with another more descriptive error and get the same behaviour? I'm assuming no for the second question.

Another avenue would be to add a new error to data capture for the same behaviour, that is more descriptive that is suitable for the LoRaWan case.

oliviamiller: We could wrap the error and display this: `ConnectError: [unknown] no readings available yet: rpc error: code = FailedPrecondition desc = no capture from filter module`. Wouldn't want to add a log line because if data capture/control page is calling readings frequently the log will spam. Open to the idea of adding a new error specific to lorawan, it would make the error more clear.

randhid: Did wrapping the error work and keep the behaviour where nothing was synced if no data? That's kinda cool if so.

gateway/join.go
```diff
@@ -34,7 +35,7 @@ type joinRequest struct {
 
 const (
 	joinRx2WindowSec = 6         // rx2 delay for sending join accept message.
-	rx2Frequenecy    = 923300000 // Frequuency for rx2 window
+	rx2Frequenecy    = 923300000 // Frequency to send downlinks on rx2 winddow
```
martha-johnston, line 38:
```suggestion
	rx2Frequenecy    = 923300000 // Frequency to send downlinks on rx2 window
```

node/node.go
oliviamiller: yes! tested it out and `errors.Is` works with wrapped error

node/node.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+type Node struct {
+	resource.Named
+	resource.AlwaysRebuild
+	logger logging.Logger
+
+	nwkSKey []byte
+	AppSKey []byte
+	AppKey  []byte
```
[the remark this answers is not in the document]
JohnN193: oh it might need to be bytes in order to send over the DoCommand

gateway/sensor.go
```diff
@@ -307,6 +187,54 @@ func (g *Gateway) updateReadings(name string, newReadings map[string]interface{}
...
+	// Validate that the dependency is correct.
+	if _, ok := cmd["validate"]; ok {
+		return map[string]interface{}{"validate": 1}, nil
+	}
+
+	// Add the nodes to the list of devices.
+	if newNode, ok := cmd["register_device"]; ok {
+		if newN, ok := newNode.(map[string]interface{}); ok {
+			node := convertToNode(newN)
+			g.devices[node.NodeName] = node
```
JohnN193, line 202: any concern if a NodeName gets renamed? we would end up with dead/unresponsive nodes right?

node/node.go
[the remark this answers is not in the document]
oliviamiller: yeah I had to change it because the type errored in docommand.

gateway/sensor.go
oliviamiller: Added another do command to delete the node on close. Then a new one will be created with new name on rebuild.

JohnN193 (approved):

gateway/join.go
```diff
@@ -133,17 +134,17 @@ func (g *Gateway) parseJoinRequestPacket(payload []byte) (joinRequest, *Device,
...
 // | 1 B  |     3 B    |   3 B |     4 B     | 1B |    1B    |  0 or 16   | 4 B  |
 // https://lora-alliance.org/wp-content/uploads/2020/11/lorawan1.0.3.pdf page 35 for more info on join accept.
-func (d *Device) generateJoinAccept(ctx context.Context, jr joinRequest) ([]byte, error) {
+func generateJoinAccept(ctx context.Context, jr joinRequest, d *node.Node) ([]byte, error) {
 	// generate random join nonce.
 	jn := generateJoinNonce()
 
 	// generate a random device address to identify uplinks.
-	d.addr = generateDevAddr()
+	d.Addr = generateDevAddr()
```
randhid, line 142: Would these need mutex protection when a node is reconfiguring? Do we anticipate an address mismatch ending up in a bad state where a node is on the old address but the gateway has rebuilt itself?

gateway/sensor.go
```diff
@@ -307,6 +187,85 @@ func (g *Gateway) updateReadings(name string, newReadings map[string]interface{}
...
+		if n, ok := name.(string); ok {
+			delete(g.devices, n)
+			g.readingsMu.Lock()
+			delete(g.lastReadings, n)
+			g.readingsMu.Unlock()
+		}
+
+	}
+
+	return nil, nil
```
randhid, line 220: return empty map[string]interface{} with nill error so that an accessor doesn't panic when trying to use the first return.

gateway/sensor.go
```diff
@@ -307,6 +187,85 @@ func (g *Gateway) updateReadings(name string, newReadings map[string]interface{}
...
+func convertToBytes(key interface{}) ([]byte, error) {
+	bytes, ok := key.([]interface{})
+	if !ok {
+		return nil, errors.New("expected node map val to be type []interface{}, but it wasn't")
+	}
+	res := make([]byte, 0)
+
+	if len(bytes) > 0 {
+		for _, b := range bytes {
+			val, _ := b.(float64)
```
randhid, line 261: check if ok?

gpio/gpioreset.go
```diff
@@ -22,18 +18,27 @@ func pinctrlSet(pin string, state string) {
 	}
 }
 
-func InitGPIO() {
+func InitGateway(resetPin, powerPin int) {
+	rst := strconv.Itoa(resetPin)
+	pwr := strconv.Itoa(powerPin)
+	initGPIO(rst, pwr)
+	resetGPIO(rst)
```
randhid, line 25: Any interestign thing about using pinctrl directly rather than our board implementations is that we don't let you map pins to names, not worried about it. It might be a feature request in the future since they are strings in your config. We'd have to add a board dependency anyway.

Does pinctrl behave as expected with physical pin numbers?

main.go
```diff
@@ -25,6 +26,10 @@ func mainWithArgs(ctx context.Context, args []string, logger logging.Logger) err
 		return err
 	}
 
+	if err = module.AddModelFromRegistry(ctx, sensor.API, node.Model); err != nil {
+		return err
+	}
+
```
randhid, line 32: Consider using ModularMain in your main.go file;

node/node.go
```diff
@@ -0,0 +1,310 @@
...
+		return nil, err
+	}
+
+	n.gateway = gateway
+
+	return n, nil
+}
+
+func (n *Node) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	cfg, err := resource.NativeConfig[*Config](conf)
```
randhid, line 189: Can you just call the  reconfigure function in the constructor, or have this resource be always rebuild, they seem to be doing the same thing in terms of populating the gateway struct?

node/node.go
```diff
@@ -0,0 +1,310 @@
...
+	var dep resource.Resource
+
+	// Assuming there's only one dep.
+	for _, val := range deps {
+		dep = val
+	}
+
+	gateway, ok := dep.(sensor.Sensor)
+	if !ok {
+		return nil, errors.New("dependency must be the sx1302-gateway")
```
randhid, line 265:
```suggestion
		return nil, errors.New("dependency must be the sx1302-gateway sensor")
```

node/node.go
```diff
@@ -0,0 +1,310 @@
...
+
+	// Validate that the dependency is the gateway - gateway will return 1.
+	ret, err := gateway.DoCommand(ctx, cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	retVal, ok := ret["validate"]
+	if !ok {
+		return nil, errors.New("dependency must be the sx1302-gateway")
```
randhid, line 279:
```suggestion
		return nil, errors.New("dependency must be the sx1302-gateway sensor")
```

node/node.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	if retVal.(float64) != 1 {
+		return nil, errors.New("dependency must be the sx1302-gateway")
```
randhid, line 282:
```suggestion
		return nil, errors.New("dependency must be the sx1302-gateway sensor")
```

node/node.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	return gateway, nil
+}
+
+func (n *Node) Close(ctx context.Context) error {
+	cmd := make(map[string]interface{})
+	cmd["remove_device"] = n.NodeName
+	n.gateway.DoCommand(ctx, cmd)
```
randhid, line 290: log the error even if we don't return it.

randhid (changes requested):
A few nits and questions, looking good. The biggest change I'd like is defaulting to OTAA in the validation logic as well, since I believe that's what the scope doc specifed.

gateway/sensor.go
```diff
@@ -67,73 +45,15 @@ func init() {
...
-	if conf.AppSKey == "" {
-		return nil, resource.NewConfigValidationError(path,
-			errors.New("app session key is required for ABP join type"))
-	}
-	if len(conf.AppSKey) != 32 {
-		return nil, resource.NewConfigValidationError(path,
-			errors.New("app session key must be 16 bytes"))
-	}
-	if conf.NwkSKey == "" {
+	if conf.ResetPin == 0 {
```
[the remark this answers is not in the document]
JohnN193: yeah, if you make the field a pointer you can differentiate between 0 and unset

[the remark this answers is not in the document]
JohnN193: cartographer does it [with bools](https://github.com/viam-modules/viam-cartographer/blob/23fefb5cc254c5b9452086bf780b945edff1a997/config/config.go#L26), but it should also work with ints

node/node.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	switch conf.JoinType {
+	case "ABP":
+		return conf.validateABPAttributes(path)
+	case "OTAA", "":
+		return conf.validateOTAAAttributes(path)
+	default:
+		return nil, resource.NewConfigValidationError(path,
+			errors.New("join type is OTAA or ABP - defaults to OTAA"))
+	}
```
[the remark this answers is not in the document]
randhid: Ah I see, my bad. Reoslving.

gpio/gpioreset.go
randhid: So does that mean on a pi we're expecting 23 to mean gpio23 - physical pin 16 on a raspberry pi?

We can change the implementation after that, but we should consider a board dependency in the future if we need to use this module with a non-Raspberry Pi board.

oliviamiller: yeah thats correct. A board dependency would make sense, can add that as a future improvement.

gateway/sensor.go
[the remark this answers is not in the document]
oliviamiller: thank you will do that!

node/node.go
oliviamiller: oops yeah I meant to remove this, I want to just do always rebuild

main.go
```diff
@@ -2,40 +2,20 @@
...
-	}
-
-	<-ctx.Done()
-	return nil
+func main() {
+	module.ModularMain(
+		moduleName,
+		resource.APIModel{API: sensor.API, Model: gateway.Model},
+		resource.APIModel{API: sensor.API, Model: node.Model},
+	)
```
randhid, line 20: This function signature has changed recently, update your rdk to update it.

gpio/gpioreset.go
randhid: Okay, we'll see how much hardware we have to accommodate other than a Raspberry Pi in the future. Adding a dependency is breaking in the future, but users could pin to the version that does not require a configuration change.

gateway/join.go
oliviamiller: thanks for pointing this out, I added reconfigure to gateway so it doesn't lose the device map if it rebuilds. Same with reconfiguring the node so that the dev addr can be saved from the gateway list if the node is reconfigured. This way the join process does not need to be redone if the node is reconfigured.

oliviamiller:
Figured out the C errors I was seeing, need to be really careful about starting/stopping the gateway because stopping a gateway that isn't running and vice versa causes hardware issues. Implemented reconfigure on the gateway to keep this from happening.

gateway/join.go
randhid: Finally, a use-case for reconfigure that I don't hate!

 For whatever reason, we need more control over the gateway keeping track of the gateways nodes across multiple shared devices from the same process running a module -; we can keep track of them in a global variable, but I don't think that is something we want to ever try to do anyway .

oliviamiller:
> > Figured out the C errors I was seeing, need to be really careful about starting/stopping the gateway because stopping a gateway that isn't running and vice versa causes hardware issues. Implemented reconfigure on the gateway to keep this from happening.
> 
> how does reconfigure get around this? just from the description it sounded like the gateway's close wasn't closing the C code & similar cases

sorry worded it poorly, you're right reconfigure doesn't get around it. It was working before with AlwaysRebuild because the gateway close method closes the C code, but with reconfigure we have to do different steps if its the first time we are starting the gateway vs reconfiguring, so added a boolean `started` to check if the gateway was started previously and close it if so.

gateway/sensor.go
```diff
@@ -67,75 +45,17 @@ func init() {
...
 	}
-	if conf.AppKey == "" {
+	if conf.Bus != 0 && conf.Bus != 1 {
 		return nil, resource.NewConfigValidationError(path,
-			errors.New("app key is required for OTAA join type"))
-	}
-	if len(conf.AppKey) != 32 {
-		return nil, resource.NewConfigValidationError(path,
-			errors.New("app key must be 16 bytes"))
+			errors.New("spi bus can be 0 or 1 - default 0"))
```
martha-johnston, line 54: in the constructor or reconfigure, is the spi bus set to 0 if it's not 0 or 1? or is this just to inform that you can set it to 0 if you aren't sure what it should be?

randhid (approved):
🥳

node/node.go
```diff
@@ -0,0 +1,263 @@
...
+func newNode(
+	ctx context.Context,
+	deps resource.Dependencies,
+	conf resource.Config,
+	logger logging.Logger,
+) (sensor.Sensor, error) {
+	n := &Node{
+		Named:    conf.ResourceName().AsNamed(),
+		logger:   logger,
+		NodeName: conf.ResourceName().AsNamed().Name().Name,
```
randhid, line 130: [q] why not the short name?

martha-johnston (approved):

PR #3, opened 2024-11-18
RSDK-9211 Add expected uplink interval to config, add timestamp to readings

oliviamiller:
Warns if the capture frequency is more than the expected interval in the config to avoid duplicate data. Add timestamp to the sensor readings.


gateway/sensor.go
```diff
@@ -196,10 +196,11 @@ func (g *Gateway) handlePacket(ctx context.Context, payload []byte) {
 			g.logger.Infof("received data uplink")
 			name, readings, err := g.parseDataUplink(ctx, payload)
 			if err != nil {
+				// don't update readings or log if from unknown device
+				if errors.Is(errNoDevice, err) {
```
JohnN193, line 200: sorry I know you talked through this already, but when would this error occur?

node/node.go
```diff
@@ -199,6 +204,25 @@ func (n *Node) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
...
+	captureFreq, err := getCaptureFrequencyHzFromConfig(conf)
+	if err != nil {
+		return nil
+	}
+	inter, err := strconv.Atoi(cfg.Interval)
+	if err != nil {
+		return nil
+	}
+	intervalSeconds := (time.Duration(inter) * time.Minute).Seconds()
+	expectedFreq := 1 / intervalSeconds
```
JohnN193, line 217: is `Interval` required? also need to make sure its nonzero to avoid divide by 0

if its not required then we should only do the freq check if we were given an interval

JohnN193 (approved):
LGTM. just some questions about how Interval is being used

node/node.go
```diff
@@ -18,6 +20,7 @@ var Model = resource.NewModel("viam", "lorawan", "node")
 type Config struct {
 	JoinType    string `json:"join_type,omitempty"`
 	DecoderPath string `json:"decoder_path"`
+	Interval    string `json:"uplink_interval_mins"`
```
JohnN193, line 23: actually why not make this a float? seems weird as a string

oliviamiller: fine with changing to a float but its a string in the scope

randhid: I'm okay with changing a type of attribute from the scope doc if it makes more sense, + there was no big motivating reason to have it be a string anyway.

oliviamiller: sounds good will change!

oliviamiller:
> It is time to write some tests for validation and Reconfigure, to make sure your creation logic doesn't break as you add new features to this module.

added tests for validate node and gateway and node construction, testing gateway constructor is a lot more complex due to C code and interaction with hardware, I can add that test in a separate PR.

node/node_test.go
```diff
@@ -0,0 +1,303 @@
...
+	_, err = conf.Validate("")
+	test.That(t, err, test.ShouldBeNil)
+}
+
+func TestNewNode(t *testing.T) {
+	ctx := context.Background()
+	logger := logging.NewTestLogger(t)
+
+	// Create mock gateway dependency
+	mockGateway := &inject.Sensor{}
```
JohnN193, line 242: can we add a GetReadings inject to this to make sure the node can get the correct reading?

gateway/gateway_test.go
```diff
@@ -28,15 +29,13 @@ func TestValidate(t *testing.T) {
...
 	conf = &Config{
 		ResetPin: &resetPin,
 		Bus:      2,
 	}
 	_, err = conf.Validate("")
-	test.That(t, err, test.ShouldNotBeNil)
-	test.That(t, err.Error(), test.ShouldContainSubstring, "spi bus can be 0 or 1")
-}
+	test.That(t, err, test.ShouldBeError, resource.NewConfigValidationError("", errInvalidSpiBus))
+}
```
randhid, line 41:
```suggestion
}

```

gateway/join.go
```diff
@@ -291,4 +284,4 @@ func reverseByteArray(arr []byte) []byte {
 		reversed[i] = arr[j]
 	}
 	return reversed
-}
+}
```
randhid, line 287:
```suggestion
}

```

gateway/sensor.go
```diff
@@ -97,17 +107,11 @@ func newGateway(
...
 	cfg, err := resource.NativeConfig[*Config](conf)
 	if err != nil {
 		return err
 	}
 
-	// If the gateway hardware was already started, stop gateway and the background worker.
-	// Make sure to always call stopGateway() before making any changes to the c config or
-	// errors will occur.
-	// Unexpected behavior will also occur if you call stopGateway() when the gateway hasn't been
-	// started, so only call stopGateway if this module already started the gateway.
```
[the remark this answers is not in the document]
oliviamiller: oops fixed

martha-johnston (approved):

PR #4, opened 2024-11-18
convert types from decoder map into protobuf compatible types 

oliviamiller:
the[ em310 tilt sensor](https://www.milesight.com/iot/product/lorawan-sensor/em310-tilt) returns some readings as a uint8, so calling readings errored because uint8 is not supported in protobuf. Adding function to convert the type to 32 bit if any 8 or 16 bit integers are returned from the decoder.

### DIFF
--- a/gateway/uplink.go
+++ b/gateway/uplink.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"gateway/node"
 	"os"
+	"reflect"
 	"time"
 
 	"github.com/robertkrimen/otto"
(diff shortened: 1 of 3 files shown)


gateway/uplink.go
```diff
@@ -63,9 +64,33 @@ func (g *Gateway) parseDataUplink(ctx context.Context, phyPayload []byte) (strin
 		return "", map[string]interface{}{}, fmt.Errorf("Error decoding payload: %w", err)
 	}
 
+	// Ensure all types in map are protobuf compatiable.
+	readings = convertTo32Bit(readings)
```
JohnN193, line 68: can we add a quick test for this?

randhid: ➕

randhid (approved):

PR #5, opened 2024-11-19
RSDK-9209 Add channel frequency list to join accept message 

oliviamiller:
This adds a channel mask to the join accept message to only enable channels 0-7, the same channels that the gateway's IF chain's are configured to receive from. This ensures that the gateway won't miss any uplinks from devices that can send from more than eight frequency channels. 

Tested manually by printing out the frequency uplinks were being sent at.


JohnN193 (approved):
assuming this does what you want, do we have a sensor we can test this with?

oliviamiller:
> assuming this does what you want, do we have a sensor we can test this with?

yep tested with all 3 lora sensors ive been testing with and works for all.

PR #7, opened 2024-11-22
Add a check before getting the framepayload

oliviamiller:
Ran into a panic when one of the devices sent a packet with no frame payload,  the framePayload will index phyPayload[9:8] and panic. Adding a check to just error and prevent the panic. 

Also adding meta.json

### DIFF
--- a/gateway/uplink.go
+++ b/gateway/uplink.go
@@ -49,7 +49,7 @@ func (g *Gateway) parseDataUplink(ctx context.Context, phyPayload []byte) (strin
 
 	// Ensure there is a frame payload in the packet.
 	if int(8+foptsLength+1) >= (len(phyPayload) - 4) {
-		return "", map[string]interface{}{}, errors.New("device sent packet with no data")
+		return "", map[string]interface{}{}, fmt.Errorf("device %s sent packet with no data", device.NodeName)
 	}
 
 	// framepayload is the device readings.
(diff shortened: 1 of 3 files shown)


gateway/uplink.go
```diff
@@ -47,7 +47,12 @@ func (g *Gateway) parseDataUplink(ctx context.Context, phyPayload []byte) (strin
 	// frame port specifies application port - 0 is for MAC commands 1-255 for device messages.
 	fPort := phyPayload[8+foptsLength]
 
-	// device data in the message.
+	// Ensure there is a frame payload in the packet.
+	if int(8+foptsLength+1) >= (len(phyPayload) - 4) {
+		return "", map[string]interface{}{}, errors.New("device sent packet with no data")
```
JohnN193, line 52: [opt] maybe add device name, though after talking with olivia it sounded like there wasn't much a user could do for this

JohnN193 (approved):

gateway/uplink.go
oliviamiller: added the name, not really anything the user can do to fix this but still may be helpful to know which device is doing this.

meta.json
```diff
@@ -0,0 +1,18 @@
...
+        {
+            "api": "rdk:component:sensor",
+            "model": "viam:lorawan:sx1302-gateway"
+         },
+         {
+            "api": "rdk:component:sensor",
+            "model": "viam:lorawan:node"
+         }
+    ],
+    "entrypoint": "mymod"
```
randhid, line 17: might want to make this something more standard in the future. Not for this pr.

randhid (approved):

meta.json
oliviamiller: yep will change when I do the CI builds.

PR #9, opened 2024-11-25
RSDK-9251 fixes to support bullseye

oliviamiller:
the `GLIBC_2.34' not found` errors are fixed by building with the antique canon version. This build works with both bullseye and bookworm. 

Using raspi-gpio instead of pinctrl if the pi is on bullseye. Wasn't able to fully test this because I don't have a bullseye pi with me while working remotely, but tested with a pi in the office that the gpios do what they're supposed to, and then got to the point where the module errored because there was no gateway hardware actually connected. 

This is just a short term solution to get bullseye working, for the long term we should have a board dependency on the gateway and call the `setGPIO` API to do the steps in `gpioreset.go`


randhid (approved):
Okay - we'll need to more generally address the builds - perhaps by packing the `.a` files from the lorawan gateway with the makefile the same way eliot has done with his pr, can you make an rc and test it with eliot's home pi?

Just moved his to rc version and it works, don't think his is bullseye though because it was building on the latest version as well.

JohnN193:
Alittle bit confused by some of the notes/comments about building, I assumed Eliot was using a version of the module on the registry and those are built before being uploaded.

There’s a few examples of modules using tarballs instead of only uploading the binary(raspberry pi & pinctrl), I’m assuming that’s what Rand was referring to with the Makefile comment but figured I would call it out

also saw we made the module public, so are we good with testing it? Doubt anyone is gonna find it but just curious

oliviamiller:
> Alittle bit confused by some of the notes/comments about building, I assumed Eliot was using a version of the module on the registry and those are built before being uploaded.
> 
> There’s a few examples of modules using tarballs instead of only uploading the binary(raspberry pi & pinctrl), I’m assuming that’s what Rand was referring to with the Makefile comment but figured I would call it out
> 
> also saw we made the module public, so are we good with testing it? Doubt anyone is gonna find it but just curious

He was using the binary in the registry but that binary had errors when running in bullseye, we have to build the module in the antique-canon image to fix, the new rc version is built in that canon shell so will work on bullseye and bookworm. What is the advantage of uploading a tarball instead of the binary? 

Eliot made the module public over the weekend so I had to change the meta.json to match in order to upload the new rc version to the registry.

JohnN193:
> > What is the advantage of uploading a tarball instead of the binary?
> 

Oh I thought you needed to include .a files or other header files that you wanted to link, and shoving everything into a zip would let you do that.

oliviamiller:
> > > What is the advantage of uploading a tarball instead of the binary?
> 
> Oh I thought you needed to include .a files or other header files that you wanted to link, and shoving everything into a zip would let you do that.

Oh okay, right now the sx1302 is statically linked so is included in the binary, the `libc` is dynamically linked so has to be present on the system. Does the pi static link the glibc library?

JohnN193:
Uhh I don’t know/haven’t looked. I only remember Susmita having to deploy her own docker container to handle the builds

JohnN193 (approved):
also sorry didn’t realize you were waiting for me but you good to merge

oliviamiller:
> Uhh I don’t know/haven’t looked. I only remember Susmita having to deploy her own docker container to handle the builds

okay cool I don't think we need to that but if there's a reason to in the future ill ask her about it.

randhid:
> There’s a few examples of modules using tarballs instead of only uploading the binary(raspberry pi & pinctrl), I’m assuming that’s what Rand was referring to with the Makefile comment but figured I would call it out

Since we have the static library files as `.a` files in this module - we could package those statically.

But viam-rdk-antique has an older glibc dependency, which Bullseye still supports, unfortunately. I'm unsure if we need a docker container for builds, that might be the way to go, but we can just upload a release with the image that works and address the underlying ioctl call differences like olivia has done here.

PR #11, opened 2024-11-26
Apply linter changes 

oliviamiller:


gateway/sensor.go
```diff
@@ -188,16 +191,15 @@ func (g *Gateway) receivePackets() {
...
-					for i := 0; i < int(packet.size); i++ {
-						payload = append(payload, byte(packet.payload[i]))
-					}
-					g.handlePacket(ctx, payload)
+				if packet.size == 0 {
+					continue
 				}
+				// Convert packet to go byte array
+				//nolint:intrange
+				for i := 0; i < int(packet.size); i++ {
```
randhid, line 199: you can do i := range packet.size in go 1.23 i believe

gateway/uplink.go
```diff
@@ -181,7 +183,7 @@ func executeDecoder(ctx context.Context, script string, vars map[string]interfac
 	}
 
 	resultChan := make(chan result)
-	timeoutCtx, _ := context.WithTimeout(ctx, 10*time.Millisecond)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
```
randhid, line 186: let's make sure this new behaviour works as expected.

randhid (approved):
manually test the behaviour changes and good to go

gateway/sensor.go
```diff
@@ -188,16 +191,15 @@ func (g *Gateway) receivePackets() {
...
-					// Convert packet to go byte array
-					for i := 0; i < int(packet.size); i++ {
-						payload = append(payload, byte(packet.payload[i]))
-					}
-					g.handlePacket(ctx, payload)
+				if packet.size == 0 {
+					continue
 				}
+				// Convert packet to go byte array
+				//nolint:intrange
```
martha-johnston, line 198: you can also add this to the `disable` section of your `etc/.golangci.yaml` file

gateway/uplink.go
```diff
@@ -119,6 +118,7 @@ func matchDeviceAddr(devAddr []byte, devices map[string]*node.Node) (*node.Node,
 }
 
 func decodePayload(ctx context.Context, fPort uint8, path string, data []byte) (map[string]interface{}, error) {
+	//nolint:gosec
```
martha-johnston, line 121: same here, can add to `disable` section, but it's up to your discretion if you think this case is an exception, or if you'd always want to disable this. when I was modularizing the builtins, if I ran into lint issues like this I would check if it was disabled in the rdk `golangci.yaml` file, and if so I would disable it in the module as well

gateway/uplink.go
```diff
@@ -192,15 +194,15 @@ func executeDecoder(ctx context.Context, script string, vars map[string]interfac
 	select {
 	case <-timeoutCtx.Done():
 		vm.Interrupt <- func() {
-			errors.New("ctx canceled")
 		}
+		cancel()
```
martha-johnston, line 198: maybe can `defer cancel()` above this instead? unless there's a situation in which you wouldn't want to call cancel before this function returns

martha-johnston (approved):

gateway/sensor.go
oliviamiller: I tried that and it errored

gateway/uplink.go
oliviamiller: I will disable this, thanks!

gateway/sensor.go
randhid: when cast to an int as well?

oliviamiller: tried it again visual studio shows it as an error but it still builds and works so changed it!

gateway/uplink.go
oliviamiller: tested that module still works correctly with all these changes!

etc/.golangci.yaml
```diff
@@ -46,6 +46,7 @@ linters:
     - varnamelen
     - wrapcheck
     - wsl
+    - gosec
```
JohnN193, line 49: Don’t disable all of gosec, just the one or two codes that you have to have for the implementation. rdk+other modules with linters should have an example

gateway/sensor.go
```diff
@@ -374,21 +377,26 @@ func convertToBytes(key interface{}) ([]byte, error) {
...
+		err := gpio.ResetGPIO(g.rstPin, *g.bookworm)
+		if err != nil {
+			g.logger.Error("error reseting gateway")
+		}
 	}
 
 	if g.workers != nil {
 		g.workers.Stop()
 	}
 	errCode := C.stopGateway()
```
JohnN193, line 392: [q]Would printing the error code be useful for debugging

JohnN193 (approved):
Just a heads up I’m off until next Tuesday, so I probably won’t be responsive on these PRs!

gateway/sensor.go
oliviamiller: no its always 1

oliviamiller: I have a ticket to better handle and surface the errors from the C code though so may readdress this there

gateway/sensor.go
randhid: ah! your vscode has older golang in its environment.

PR #12, opened 2024-11-26
RSDK-9350 turn on i2c and spi on first run

oliviamiller:
Use the first run script to turn on i2c and spi on the pi.

Tested the make file command changes locally `make lorawan` and `make module.tar.gz`, and tested that script can successfully turn on the protocols.


martha-johnston (approved):

Makefile
```diff
@@ -5,16 +5,15 @@ BIN_OUTPUT_PATH = bin
...
-	CGO_LDFLAGS="$$CGO_LDFLAGS $(CGO_BUILD_LDFLAGS)" go build $(GO_BUILD_LDFLAGS) -o  $(BIN_OUTPUT_PATH)/lorawan main.go
+lorawan:
+	rm -f lorawan
+	CGO_LDFLAGS="$$CGO_LDFLAGS $(CGO_BUILD_LDFLAGS)" go build $(GO_BUILD_LDFLAGS) -o $@ main.go
 
-module.tar.gz: sx1302 build
+module.tar.gz: lorawan first_run.sh meta.json
 	rm -f $(BIN_OUTPUT_PATH)/module.tar.gz
-	tar czf $(BIN_OUTPUT_PATH)/module.tar.gz $(BIN_OUTPUT_PATH)/lorawan
+	tar czf $(BIN_OUTPUT_PATH)/module.tar.gz $^
```
martha-johnston, line 14: unfamiliar with this ^ format, but I assume it puts the build one level up from bin, since bin was removed from the entry point file path?

first_run.sh
```diff
@@ -0,0 +1,32 @@
...
+    local protocol=$1
+    local get_cmd="get_${protocol}"
+    local do_cmd="do_${protocol}"
+
+    # Check the current status of the protocol
+    status=$(sudo raspi-config nonint "$get_cmd")
+    if [[ $? -ne 0 ]]; then
+        echo "Failed to get ${protocol} status. Ensure that raspi-config is installed."
+        exit 1
+    fi
```
randhid, line 14: Check that you're on a rapsberry pi first.

Also I notice you're using bin/bash. I think the syntax is correct, I'm expecting all the checks to be in `[[ ]]` but just run this file on a random raspberry pi after you turn off i2c to ensure it works.

Makefile
```diff
@@ -5,16 +5,15 @@ BIN_OUTPUT_PATH = bin
...
-.PHONY: build
-build:
-	rm -f $(BIN_OUTPUT_PATH)/lorawan
-	CGO_LDFLAGS="$$CGO_LDFLAGS $(CGO_BUILD_LDFLAGS)" go build $(GO_BUILD_LDFLAGS) -o  $(BIN_OUTPUT_PATH)/lorawan main.go
+lorawan:
+	rm -f lorawan
+	CGO_LDFLAGS="$$CGO_LDFLAGS $(CGO_BUILD_LDFLAGS)" go build $(GO_BUILD_LDFLAGS) -o $@ main.go
 
-module.tar.gz: sx1302 build
+module.tar.gz: lorawan first_run.sh meta.json
```
randhid, line 12: let's not mess with the makefile too much every pr going forward.

randhid (changes requested):
Check that you're on a pi first.

Makefile
oliviamiller: the module.tar.gz has to include these things for the first_run to work but yeah this should be the last PR with makefile changes

Makefile
oliviamiller: the `$^` tells it to include all the prereq files into the tar. So now instead of the tar just being the module binary we're bundling the binary, `first_run.sh`, and the `meta.json` into the tar. the `lorawan` binary is now put one level up from bin and the `module.tar.gz` is still put in bin.

first_run.sh
oliviamiller: Added check for rpi, tested on my pi and it works as expected!

randhid (approved):

PR #13, opened 2024-12-03
RSDK-9303 Add unit tests for gateway logic

oliviamiller:
Added unit tests for functions in `join.go`, `uplink.go`, and `gateway.go` that don't have hardware dependencies.


gateway/gateway_test.go
```diff
@@ -39,3 +42,198 @@ func TestValidate(t *testing.T) {
 	_, err = conf.Validate("")
 	test.That(t, err, test.ShouldBeError, resource.NewConfigValidationError("", errInvalidSpiBus))
 }
+
+func TestDoCommand(t *testing.T) {
+	g := &Gateway{
+		devices: make(map[string]*node.Node),
+	}
```
randhid, line 49: [q] Is there a reason the gateway is exported? Okay to leave if yes.

gateway/gateway_test.go
```diff
@@ -39,3 +42,198 @@ func TestValidate(t *testing.T) {
...
+	test.That(t, ok, test.ShouldBeTrue)
+	test.That(t, dev.AppKey, test.ShouldResemble, testAppKey)
+	test.That(t, dev.DevEui, test.ShouldResemble, testDevEUI)
+	test.That(t, dev.DecoderPath, test.ShouldResemble, "/newpath")
+
+	// Test remove device command
+	removeCmd := make(map[string]interface{})
+	removeCmd["remove_device"] = n.NodeName
+	doOverWire(s, removeCmd)
+	test.That(t, len(g.devices), test.ShouldEqual, 0)
```
randhid, line 104: Can we test some error cases of your DoCommands, if there are any in this testing function?

gateway/gateway_test.go
```diff
@@ -39,3 +42,198 @@ func TestValidate(t *testing.T) {
...
+		"hum":  60,
+	}
+	g.updateReadings("device1", newReadings)
+	test.That(t, g.lastReadings["device1"], test.ShouldResemble, newReadings)
+
+	// Test updating existing readings
+	updatedReadings := map[string]interface{}{
+		"temp": 26.5,
+		"pres": 1013,
+	}
```
randhid, line 213: This might benefit from a test case structure too.
```suggestion
	expected := map[string]interface{}{
		"temp": 26.5,
		"pres": 1013,
	}
```

gateway/gateway_test.go
```diff
@@ -39,3 +42,198 @@ func TestValidate(t *testing.T) {
...
+			"device2": map[string]interface{}{
+				"pres": 1013,
+			},
+		},
+	}
+
+	readings, err := g.Readings(context.Background(), nil)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, readings, test.ShouldResemble, g.lastReadings)
+}
```
randhid, line 239: Can we combine with the test above, this isn't super useful on it's own for testing any functionality. and updateReadings -> expected readings from the API call is a better test of [encapsulation](https://testmatick.com/encapsulation-as-one-of-the-fundamental-principles-of-object-oriented-programming/).

gateway/join_test.go
```diff
@@ -0,0 +1,206 @@
+package gateway
+
+import (
+	"context"
```
randhid, line 4: These are super nice tests for your helpers.

randhid (commented):
Few nits, but looking good overall, do we have a sense of how much coverage we have when we run go test?  Also add a description to this PR with what was added, do that for every pr you make.

gateway/join_test.go
```diff
@@ -0,0 +1,206 @@
...
+			test.That(t, result, test.ShouldResemble, tt.expected)
+		})
+	}
+}
+
+// test that random dev addr is 4 bytes and 7 msb is network id.
+func TestGenerateDevAddr(t *testing.T) {
+	addr := generateDevAddr()
+	test.That(t, len(addr), test.ShouldEqual, 4)
+	test.That(t, addr[0], test.ShouldEqual, netID[0])
```
martha-johnston, line 64: where are `netID[0]` and `netID[1]` defined?

gateway/join_test.go
```diff
@@ -0,0 +1,206 @@
...
+	joinAccept, err := generateJoinAccept(ctx, jr, device)
+	test.That(t, err, test.ShouldBeNil)
+	// MHDR(1) + Encrypted(JoinNonce(3) + NetID(3) + DevAddr(4) + DLSettings(1) + RxDelay(1) + CFList(16)) + MIC(4) = 33 bytes
+	test.That(t, len(joinAccept), test.ShouldEqual, 33)
+	// Join-accept message type
+	test.That(t, joinAccept[0], test.ShouldEqual, byte(0x20))
+	// Device address should be generated and added to device.
+	test.That(t, len(device.Addr), test.ShouldEqual, 4)
+	// AppSKey should be generated and added to device
+	test.That(t, len(device.AppSKey), test.ShouldEqual, 16)
```
martha-johnston, line 163: maybe consider changing the expected numbers here to variables

gateway/join_test.go
```diff
@@ -0,0 +1,206 @@
...
+	payload := []byte{0x00} // MHDR
+	payload = append(payload, testJoinEUI...)
+	payload = append(payload, testDevEUILE...)
+	payload = append(payload, testDevNonce...)
+
+	mic, err := crypto.ComputeJoinRequestMIC(types.AES128Key(testDevice.AppKey), payload)
+	test.That(t, err, test.ShouldBeNil)
+	payload = append(payload, mic[:]...)
+
+	// Test with context that will timeout before rx2 window
```
martha-johnston, line 190: what's the point of this timing out if it's not to test an error?

gateway/mockdecoder.js
```diff
@@ -0,0 +1,87 @@
...
+    return { data: decoded };
+}
+
+// Chirpstack v3
+function Decode(fPort, bytes) {
+    return DeviceDecode(bytes);
+}
+
+// The Things Network
+function Decoder(bytes, port) {
```
martha-johnston, line 13: why `Decoder` and `Decode`?

martha-johnston (approved):
lgtm!

gateway/gateway_test.go
oliviamiller: the expected is defined below this, maybe `newReading` if `updatedReadings` is confusing?

gateway/join_test.go
oliviamiller: `join.go`

gateway/mockdecoder.js
oliviamiller: Decoder is used by the things network and decode is used by chirpstack, I just modeled the mock decoder exactly how a real one looks like.

oliviamiller: (things network and chirpstack are both popular lorawan network servers)

gateway/join_test.go
martha-johnston: so generateDevAddr generates the same address every time?

gateway/join_test.go
martha-johnston: I think it's fine to leave. if things change in the future you might want to test this functionality specifically

gateway/gateway_test.go
oliviamiller: I can make it private

gateway/gateway_test.go
randhid: How about a structure that looks like this so you're testing an API and the update readings fucntion, right not I guess it's okay since it's an input and output test but we can add more coverage within this test by perhaps doing:
```go
expected := some readings
g.updateReadings("device", expected)
readings, err := g.Readings(....)
test.That(t, err, test.ShouldBeNil)
test,That(t, readings, tets.ShouldResemble, expected)
```

Also are there any error cases we expect from update readings? We should test those too.

gateway/join_test.go
oliviamiller: no, the first two bytes are always the same but the last two bytes are random.

gateway/gateway_test.go
oliviamiller: The return of Readings() doesn't always equal the input to updateReadings, but I will combine the readings and updateReadings test. Theres not error cases.

oliviamiller: Actually I would prefer to keep the tests separate, I think its a lot cleaner to keep one test per function

randhid: My  issue now is that there is no update readings -> last readings -> the actual API that is called: `Readings` test, so we would miss a case where something goes wrong where lastreadings is updated but that itsn't updated in the readings map. 

One test per function is cleaner, but make sure you're testing code paths that actually make sense to test, right now the Readings test does not tell me much about this driver's behaviour.

oliviamiller: Doesn't the `TestReadings ` below this test what you are referring to? The purpose of the Readings test below is to make sure that `g.lastReadings` and the readings map match.

PR #14, opened 2024-12-05
RSDK-9352 Improve error handling of C code

oliviamiller:
This PR redirects C's stdout to a pipe so that we can log the C errors in real time. This will make it far easier to debug if anything goes wrong in the sx1302 code. Also added debug logs for all other prints in the C code (chose to make it debug opposed to info since its pretty verbose). 

Also adds a function to parse the error codes from the` setupGateway` function, if something errors in that function the user will see the C error log followed by the error code line from the constructor.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -16,14 +16,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"gateway/gpio"
+	"gateway/node"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-
-	"gateway/gpio"
-	"gateway/node"
 
 	"go.viam.com/rdk/components/sensor"
 	"go.viam.com/rdk/logging"
@@ -94,10 +93,10 @@ type gateway struct {
 	rstPin   string
 }
 
-// global variable to track if the captureOutputToLogs goroutine has started
-// If the build errors and needs to build again, we only want to start the logging goroutine
-// once.
-var logsStarted bool = false
+// LoggingRoutineStarted is a global variable to track if the captureCOutputToLogs goroutine has
+// started for each gateway. If the gateway build errors and needs to build again, we only want to start
+// the logging routine once.
+var LoggingRoutineStarted = make(map[string]bool)
 
 // NewGateway creates a new gateway
 func NewGateway(
@@ -149,11 +148,13 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		g.started = false
 	}
 
-	// capture c log output
-	if !logsStarted {
+	// capture C log output
+	loggingStarted, ok := LoggingRoutineStarted[g.Name().Name]
+	if !ok || !loggingStarted {
 		g.workers = utils.NewBackgroundStoppableWorkers(g.captureCOutputToLogs)
-		logsStarted = true
+		LoggingRoutineStarted[g.Name().Name] = true
 	}
+
 	// maintain devices and lastReadings through reconfigure.
 	if g.devices == nil {
 		g.devices = make(map[string]*node.Node)
@@ -212,7 +213,9 @@ func (g *gateway) captureCOutputToLogs(ctx context.Context) {
 	C.redirectToPipe(C.int(stdoutW.Fd()))
 	scanner := bufio.NewScanner(stdoutR)
 
+	//nolint:errcheck
 	defer stdoutR.Close()
+	//nolint:errcheck
 	defer stdoutW.Close()
 
 	// loop to read lines from the scanner and log them
@@ -448,7 +451,7 @@ func (g *gateway) Close(ctx context.Context) error {
 	}
 	if g.workers != nil {
 		g.workers.Stop()
-		logsStarted = false
+		LoggingRoutineStarted[g.Name().Name] = false
 	}
 	errCode := C.stopGateway()
 	if errCode != 0 {
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -234,9 +234,10 @@ func (g *gateway) captureCOutputToLogs(ctx context.Context) {
 		default:
 			if scanner.Scan() {
 				line := scanner.Text()
-				if strings.Contains(line, "ERROR") {
+				switch {
+				case strings.Contains(line, "ERROR"):
 					g.logger.Error(line)
-				} else {
+				default:
 					g.logger.Debug(line)
 				}
 			}
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -16,13 +16,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"gateway/gpio"
-	"gateway/node"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"gateway/gpio"
+	"gateway/node"
 
 	"go.viam.com/rdk/components/sensor"
 	"go.viam.com/rdk/logging"
@@ -47,6 +48,11 @@ var (
 
 // Model represents a lorawan gateway model.
 var Model = resource.NewModel("viam", "lorawan", "sx1302-gateway")
+
+// LoggingRoutineStarted is a global variable to track if the captureCOutputToLogs goroutine has
+// started for each gateway. If the gateway build errors and needs to build again, we only want to start
+// the logging routine once.
+var loggingRoutineStarted = make(map[string]bool)
 
 // Config describes the configuration of the gateway
 type Config struct {
@@ -92,11 +98,6 @@ type gateway struct {
 	bookworm *bool
 	rstPin   string
 }
-
-// LoggingRoutineStarted is a global variable to track if the captureCOutputToLogs goroutine has
-// started for each gateway. If the gateway build errors and needs to build again, we only want to start
-// the logging routine once.
-var LoggingRoutineStarted = make(map[string]bool)
 
 // NewGateway creates a new gateway
 func NewGateway(
@@ -149,11 +150,7 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 	}
 
 	// capture C log output
-	loggingStarted, ok := LoggingRoutineStarted[g.Name().Name]
-	if !ok || !loggingStarted {
-		g.workers = utils.NewBackgroundStoppableWorkers(g.captureCOutputToLogs)
-		LoggingRoutineStarted[g.Name().Name] = true
-	}
+	g.startCLogging(ctx)
 
 	// maintain devices and lastReadings through reconfigure.
 	if g.devices == nil {
@@ -195,6 +192,17 @@ func parseErrorCode(errCode int) string {
 		return "error starting the gateway"
 	default:
 		return "unknown error"
+	}
+}
+
+// startCLogging starts the goroutine to capture C logs into the logger.
+// If loggingRoutineStarted indicates routine has already started, it does nothing.
+func (g *gateway) startCLogging(ctx context.Context) {
+	loggingState, ok := loggingRoutineStarted[g.Name().Name]
+	if !ok || !loggingState {
+		g.logger.Debug("Starting c logger background routine")
+		g.workers = utils.NewBackgroundStoppableWorkers(g.captureCOutputToLogs)
+		loggingRoutineStarted[g.Name().Name] = true
 	}
 }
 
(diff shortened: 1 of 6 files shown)


gateway/sensor.go
```diff
@@ -171,10 +181,59 @@ func (g *Gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+		return "error setting the radio frequency config for radio 0"
+	case 3:
+		return "error setting the radio frequency config for radio 1"
+	case 4:
+		return "error setting the intermediate frequency chain config"
+	case 5:
+		return "error starting the gateway"
+	default:
+		return "unknown error"
+	}
```
JohnN193, line 198: [opt] only thought is defining these errors as constants, but I think thats only actually useful for tests or if you wanted other repos to be able to use these error codes.

gateway/sensor.go
```diff
@@ -171,10 +181,59 @@ func (g *Gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+}
+
+// captureOutput creates a goroutine to capture C's stdout and log to the module's logger.
+// This is necessary because the sx1302 library only uses printf to report errors.
+func (g *Gateway) captureOutput() error {
```
JohnN193, line 203: [opt] maybe update the name to specify its capturing logs from C? I confused myself thinking this was the output of a sensor

gateway/sensor.go
```diff
@@ -171,10 +181,59 @@ func (g *Gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+	stdoutR, stdoutW, err := os.Pipe()
+	if err != nil {
+		return err
+	}
+	// Redirect C's stdout to the write end of the pipe
+	C.redirectToPipe(C.int(stdoutW.Fd()))
+
+	g.workers.Add(func(ctx context.Context) {
```
JohnN193, line 211: [opt] rather than define a function here to add as a worker, can we just pass in this whole function? similar to how we [do it for gps](https://github.com/viam-modules/gps/blob/bfae1ce1fbeae842c949e9d4e5b6024be3329a4b/rtk/gpsrtk.go#L64)

gateway/sensor.go
```diff
@@ -106,9 +106,19 @@ func NewGateway(
...
 		logger:  logger,
 		started: false,
+		workers: utils.NewBackgroundStoppableWorkers(),
+	}
+
+	// Need to disable buffering on stdout so C logs can be displayed in real time.
+	C.disableBuffering()
+
+	// capture c log output
+	err := g.captureOutput()
```
JohnN193, line 116: [opt] to go along with the other comment, I think adding to the worker from the NewGateway function makes its clear to a reader that a background thread is being spun up. note you would technically need to add a context as an input to captureOutput

```suggestion
		started: false,
	}

	// Need to disable buffering on stdout so C logs can be displayed in real time.
	C.disableBuffering()

	// capture c log output
	g.workers =  utils.NewBackgroundStoppableWorkers(g.captureOutput)
```

gateway/sensor.go
```diff
@@ -171,10 +181,59 @@ func (g *Gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+		}
+	})
+	return nil
+}
+
 func (g *Gateway) receivePackets() {
 	// receive the radio packets
 	packet := C.createRxPacketArray()
-	g.workers = utils.NewBackgroundStoppableWorkers(func(ctx context.Context) {
+	g.workers.Add(func(ctx context.Context) {
```
JohnN193, line 236: [opt] similar comment, I think adding `receivePackets` to the worker in NewGateway would make it clearer to a reader that a background thread is getting spun up here.

JohnN193 (approved):
LGTM. basically just wondering if we should be adding threads to the worker more explicitly in the NewGateway function

gateway/sensor.go
```diff
@@ -171,10 +181,59 @@ func (g *Gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				if scanner.Scan() {
+					line := scanner.Text()
+					if strings.Contains(line, "ERROR") {
+						g.logger.Error(line)
+					} else {
+						g.logger.Debug(line)
```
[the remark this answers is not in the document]
JohnN193: nah no need for that. I wasn't sure if the c code had some delay before it was actually ready

randhid (approved):

gateway/sensor.go
```diff
@@ -162,50 +164,98 @@ func (g *Gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
 	errCode := C.setUpGateway(C.int(cfg.Bus))
 	if errCode != 0 {
-		return fmt.Errorf("failed to start the gateway %d", errCode)
+		strErr := parseErrorCode(int(errCode))
+		return fmt.Errorf("failed to set up the gateway: %s", strErr)
 	}
 
 	g.started = true
-	g.receivePackets()
+	g.workers.Add(g.receivePackets)
```
randhid, line 172: Do we have a `workers == nil` check before we stop this routine somewhere?

oliviamiller: yes! in `Close`

gateway/sensor.go
```diff
@@ -95,6 +94,11 @@ type Gateway struct {
 	rstPin   string
 }
 
+// global variable to track if the captureOutputToLogs goroutine has started
+// If the build errors and needs to build again, we only want to start the logging goroutine
+// once.
+var logsStarted bool = false
```
randhid: if we have multiple sensors and close one is closed this will be overwritten. we might want it to be contained in the sensors struct or mapped to a sensor name.

randhid (changes requested):
the logging global variable can now be overwritten if multiple sensors are present and one is closed.

gateway/sensor.go
oliviamiller: Okay yeah I see i didn't think about having multiple thanks for pointing that out, I can't put it in the gateway struct because it gets overwritten on rebuild, I think making a map with the gateway name would work, will try to do that.

gateway/gateway.go
```diff
@@ -386,9 +449,9 @@ func (g *gateway) Close(ctx context.Context) error {
 			g.logger.Error("error reseting gateway")
 		}
 	}
-
 	if g.workers != nil {
 		g.workers.Stop()
+		LoggingRoutineStarted[g.Name().Name] = false
```
randhid, line 454: We may want to remove the element from the map entirely.

gateway/gateway.go
```diff
@@ -146,6 +148,13 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		g.started = false
 	}
 
+	// capture C log output
+	loggingStarted, ok := LoggingRoutineStarted[g.Name().Name]
+	if !ok || !loggingStarted {
+		g.workers = utils.NewBackgroundStoppableWorkers(g.captureCOutputToLogs)
+		LoggingRoutineStarted[g.Name().Name] = true
+	}
+
```
randhid, line 157: I can't add two resources with the same name from the resource graph, but I can with the exported constructor function call, can we ensure that the Logger map is overwritten in the correct way, and also make this logic into a struct with functions that are testable, and write tests for them to ensure that we append the logger map in the correct way.

gateway/gateway.go
```diff
@@ -146,6 +148,13 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
 	// maintain devices and lastReadings through reconfigure.
 	if g.devices == nil {
 		g.devices = make(map[string]*node.Node)
```
randhid, line 160: So at this point, we just sat without a clogger?

gateway/gateway.go
```diff
@@ -146,6 +148,13 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		g.started = false
 	}
 
+	// capture C log output
+	loggingStarted, ok := LoggingRoutineStarted[g.Name().Name]
+	if !ok || !loggingStarted {
```
randhid, line 153: what's the case where we're not okay, but the logging has not started?

If you want just a check for an element existing we can use a map[string]interface{} rather than a map[string]bool

gateway/gateway.go
```diff
@@ -95,6 +93,11 @@ type gateway struct {
 	rstPin   string
 }
 
+// LoggingRoutineStarted is a global variable to track if the captureCOutputToLogs goroutine has
+// started for each gateway. If the gateway build errors and needs to build again, we only want to start
+// the logging routine once.
+var LoggingRoutineStarted = make(map[string]bool)
```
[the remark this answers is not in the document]
oliviamiller: can unexport

gateway/gateway.go
oliviamiller: If the element does not exist in the map, but yeah in that case we can just change the map to a slice of names instead of a map.

gateway/gateway.go
oliviamiller: not sure I understand the question, can you elaborate

gateway/gateway.go
randhid: That is to say, we'll keep this map as it already is.

gateway/gateway_test.go
```diff
@@ -233,3 +234,33 @@ func TestUpdateReadings(t *testing.T) {
...
+	test.That(t, loggingRoutineStarted["test-gateway"], test.ShouldBeTrue)
+	test.That(t, g.workers, test.ShouldNotBeNil)
+
+	// Test case 2: Ensure no new goroutine is started if the loggingRoutineStarted entry is true.
+	// reset g.workers to test for new workers
+	g.workers = nil
+	g.startCLogging(ctx)
+	test.That(t, g.workers, test.ShouldBeNil)
+	test.That(t, loggingRoutineStarted["test-gateway"], test.ShouldBeTrue)
+
```
randhid, line 265: Let's add a test to ensure that the closure of a sensor removes the logger name from the clogging map.

gateway/gateway.go
```diff
@@ -162,50 +168,115 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+		case <-ctx.Done():
+			return
+		default:
+			if scanner.Scan() {
+				line := scanner.Text()
+				if strings.Contains(line, "ERROR") {
+					g.logger.Error(line)
+				} else {
+					g.logger.Debug(line)
+				}
```
randhid, line 241: [nit, optional] A case structure is always easier to read to me than an if-else statement.

randhid (commented):
Test the removal of the logger name and bool from the map at resource closure as well please.

gateway/gateway.go
oliviamiller: sure can change!

gateway/gateway_test.go
```diff
@@ -252,15 +251,21 @@ func TestStartCLogging(t *testing.T) {
...
 	g.startCLogging(ctx)
 	test.That(t, loggingRoutineStarted["test-gateway"], test.ShouldBeTrue)
 	test.That(t, g.workers, test.ShouldNotBeNil)
 
-	// Test case 2: Ensure no new goroutine is started if the loggingRoutineStarted entry is true.
-	// reset g.workers to test for new workers
+	// Test that closing the gateway removes the gateway from the loggingRoutineStarted map.
+	err := g.Close(ctx)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(loggingRoutineStarted), test.ShouldEqual, 0)
```
randhid, line 262: We should be equivalent in our tests - test that worker are nil in this case, and  also that that logginroutine length is 1 in the test above.

oliviamiller: `Close` doesn't set the workers as nil it just stops them, I dont think theres a way to test if they are running or stopped so I was using `ShouldBeNil` as kind of a workaround to test for that, but it wouldn't work here.

oliviamiller: I could also just add a `g.workers = nil` line to the Close function but it wouldn't be necessary for anything except the test.

randhid: Hmm, that is surprising since I assumed by using the `sync.WaitGroup`'s done function, `StoppableWorker`'s `cancel` function would decrement the workers count in its implementation in goutils. But if it does not then feel free to just check the length for parity between the two test cases.

I'll have to think a little about the `stoppableworkers.Stop` function not decrementing the worker group, for another day!

oliviamiller: I believe calling stop does decrement the worker group, but the wait group count is not exposed by go. see [this](https://github.com/golang/go/issues/7202)

randhid: Oh awesome find, also very good to know that we can't check it in that way.

Thanks for looking into this!

randhid (approved):

PR #15, opened 2024-12-16
RSDK-9494 Fix the no reading available yet error log

oliviamiller:
The `fmt.Errorf("no readings available yet: %w", data.ErrNoCaptureToStore)` started spamming logs when you were on the RC card. Making a change to only return the `data.ErrNoCaptureToStore` error when the call is from data manager, and otherwise return no error and just the empty readings map. This prevents the spam of errors when on the RC card. 

Also added the errNoCaptureToStore error to the gateway to prevent capturing empty data.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -24,11 +24,6 @@ import (
 
 	"gateway/gpio"
 	"gateway/node"
-	"os"
-	"strconv"
-	"strings"
-	"sync"
-	"time"
 
 	"go.viam.com/rdk/components/sensor"
 	"go.viam.com/rdk/data"
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -16,6 +16,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
 	"gateway/gpio"
 	"gateway/node"
 	"os"
@@ -23,6 +29,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+
 	"go.viam.com/rdk/components/sensor"
 	"go.viam.com/rdk/data"
 	"go.viam.com/rdk/logging"
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -16,16 +16,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


gateway/uplink.go
```diff
@@ -6,12 +6,10 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"gateway/node"
 	"os"
 	"reflect"
 	"time"
-
-	"gateway/node"
-
```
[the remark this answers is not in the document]
JohnN193: yeah it does that for some reason also not sure why. running the linter a 2nd time usually reverts the change

gateway/gateway.go
```diff
@@ -473,5 +472,14 @@ func (g *gateway) Close(ctx context.Context) error {
 func (g *gateway) Readings(ctx context.Context, extra map[string]interface{}) (map[string]interface{}, error) {
 	g.readingsMu.Lock()
 	defer g.readingsMu.Unlock()
+
+	// no readings available yet
+	if len(g.lastReadings) == 0 {
+		// Tell the collector not to capture the empty data.
+		if extra[data.FromDMString] == true {
```
JohnN193, line 486: wondering if we need to have an `ok` check but based on tests it seems like we don't

gateway/gateway_test.go
```diff
@@ -235,6 +235,35 @@ func TestUpdateReadings(t *testing.T) {
...
+		lastReadings: make(map[string]interface{}),
+	}
+
+	// If lastReadings is empty and the call is not from data manager, return no error.
+	readings, err := g.Readings(context.Background(), nil)
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, readings, test.ShouldResemble, map[string]interface{}{})
+
+	// If lastReadings is empty and the call is from data manager, return ErrNoCaptureToStore
+	readings, err = g.Readings(context.Background(), map[string]interface{}{data.FromDMString: true})
```
JohnN193: [opt/q] is the behavior the same if you have `data.FromDMString: false`? just curious with how interfaces work

JohnN193 (approved):

gateway/gateway_test.go
oliviamiller: no error should be returned if the FromDMString is false, added a test case for that.

PR #16, opened 2024-12-16
Add clean and default to makefile

oliviamiller:
adding a clean to remove the binary and tar and default that cleans and then builds the binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,13 @@ UNAME_S ?= $(shell uname -s)
 # flag for testing
 CGO_CFLAGS := -DTESTING
 
-lorawan: sx1302
+default: clean lorawan
+
+clean:
 	rm -f lorawan
+	rm -f module.tar.gz
+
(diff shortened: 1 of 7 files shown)


martha-johnston (approved):

Makefile
```diff
@@ -7,8 +7,13 @@ UNAME_S ?= $(shell uname -s)
 # flag for testing
 CGO_CFLAGS := -DTESTING
 
-lorawan: sx1302
+default: clean lorawan
```
JohnN193, line 10: not make the module by default? or is this better for testing

JohnN193 (approved):

Makefile
oliviamiller: Yeah I feel like this is more useful for testing since the make module will usually be done with CI.

JohnN193: gotcha def leave it with whatever is best for your workflow

PR #17, opened 2024-12-19
RSDK-9444 Add board dependency to the gateway

oliviamiller:
Added a board dependency to the gateway config, this allows us to use our APIs instead of pinctrl and will allow us to potentially support boards other than pi in the future. Had to fix a closing issue by changing the order that we reset/close routines and the gateway during reconfigure and shutdown. 

Tested on pi4 and works as expected.

To do:

- [x] test with pi5


gateway/gpio.go
```diff
@@ -0,0 +1,50 @@
+package gateway
```
oliviamiller, line 1: didn't think this code needed to be in a separate package anymore since only gateway uses it so moved here.

README.md
```diff
@@ -134,6 +135,8 @@ The gateway communicates through SPI, ensure that SPI in enabled on the pi.
 
 To avoid capturing duplicate data, set the data capture frequency equal to or less than the expected uplink interval.
 
+If the error `ERROR: Failed to set SX1250_0 in STANDBY_RC mode` is seen in logs, unplug the raspberry pi for a few minutes and then try again.
```
oliviamiller, line 138: I wouldn't expect a user to hit this once this change is in but added here just in case.

JohnN193: would it be possible for our logs to also return this suggestion?

oliviamiller: yep, added!

gateway/gpio.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+	"time"
+
+	"go.viam.com/rdk/components/board"
+	"go.viam.com/utils"
+)
+
+// it is necessary to sleep betweeen setting the gpio pins - the gateway will not initialize correctly without it.
+func waitGPIO(ctx context.Context) error {
+	if !utils.SelectContextOrWait(ctx, 100*time.Millisecond) {
```
JohnN193, line 14: [nit] can we make the 100 a constant at the top?

JohnN193 (approved):
LGTM just some nits. Mainly not sure about the lock in `g.reset()`, if we could just let that lock live in Close() and Reconfigure() I think that would be better

gateway/gateway.go
```diff
@@ -123,31 +123,39 @@ func NewGateway(
...
-		err := g.Close(ctx)
+		err := g.reset(ctx)
 		if err != nil {
 			return err
 		}
-		g.started = false
+	}
+
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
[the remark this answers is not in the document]
oliviamiller: the lock is in `receivePackets ` not reset, but yeah Im not 100% sure if it is necessary there so I can take it out.

gateway/gateway.go
```diff
@@ -123,31 +125,38 @@ func NewGateway(
...
+	var dep resource.Resource
+
+	// Assuming there's only one dep.
+	for _, val := range deps {
+		dep = val
+	}
+
+	board, ok := dep.(board.Board)
+	if !ok {
+		return errors.New("dependency must be the board")
```
martha-johnston, line 159:
```suggestion
		return errors.New("dependency must be a raspberry pi board")
```

gateway/gateway.go
```diff
@@ -123,31 +125,38 @@ func NewGateway(
...
+	}
+	if len(deps) == 0 {
+		return errors.New("must add raspberry pi board as dependency")
+	}
+	var dep resource.Resource
+
+	// Assuming there's only one dep.
+	for _, val := range deps {
+		dep = val
+	}
```
randhid, line 155: This is wrong if a user adds another dep that is also a board, it's non-deterministic because the assumption that there's only one dep is wrong. Either refuse to build without one dep or check all the dependencies to ensure that only one is a board and use that one. This, however, is confusing to troubleshoot and ripe for useability bugs if a user adds two board dependencies to their config.

Could you use the board?FromDependencies with an implicit dependency addition instead? It will break configs as it adds a required attribute. It wasn't in your scope doc either, so it's worth a call out in that document as well. We still have very low useability of this module, so changing those configs to the new version is priority number one after this version's release.

Also, please move this module code to viam-modules so that tickets can be tracked on Jira; it's already public, unfortunately, so there's no added benefit to having it in your GitHub org anymore.

oliviamiller: sure, this is already a breaking change because a user will need to add the board to the depends_on, so we might as well make it an implicit dependency with board attribute. Also want to point out that im doing the exact same thing in node.go to get the gateway dependency, so may be worth changing the gateway to be an implicit dependency before we have many users.

randhid (approved):
Explicitly LGTMing! Please talk to eliot + chris to update their configs to include a board dependency. As of 12/26/2024 there are 5 configs with lorwan modules - 3 of which are yours olivia so not too worried about the footprint of the break.

PR #19, opened 2025-01-21
RSDK-9780 fix nil readings error

oliviamiller:
Due to the recent updates in the sensor server to check for nil returns, returning an empty map will result in the error `calling handler error rpc error: code = Unknown desc = sensor component tilt Readings should not return nil readings`. To avoid this I am instead returning an empty string in the readings map stating that there are no readings yet.

Tested manually and updated unit tests.


randhid (approved):
Thanks for the quick ux improvement!

PR #22, opened 2025-01-29
RSDK-9808 Persist OTAA data across restarts

oliviamiller:
Currently, if the module is restarted the dev addr and keys will be lost and the device needs to resend a join request. This PR writes the OTAA info of each device to a file during the join procedure. When a previously known device is registered to the gateway through the docommand, the device's info is pulled from the file and added to the device object. Now the device does not need to be manually reset if the module gets restarted.

Testing: Added unit tests for each function, tested manually using tilt sensor.

To do:
[ ] more manual testing

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -401,28 +401,27 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 			}
 
 			oldNode, exists := g.devices[node.NodeName]
-			if !exists {
+			switch exists {
+			case false:
 				g.devices[node.NodeName] = node
-				return map[string]interface{}{}, nil
+			case true:
+				// node with that name already exists, merge them
+				mergedNode, err := mergeNodes(node, oldNode)
+				if err != nil {
+					return nil, err
+				}
+				g.devices[node.NodeName] = mergedNode
 			}
-			// node with that name already exists, merge them
-			mergedNode, err := mergeNodes(node, oldNode)
-			if err != nil {
-				return nil, err
-			}
-			g.devices[node.NodeName] = mergedNode
-
 			// Check if the device is in the persistent data file, if it is add the OTAA info.
-			deviceInfo, err := g.searchForDeviceInFile(mergedNode.DevEui)
+			deviceInfo, err := g.searchForDeviceInFile(g.devices[node.NodeName].DevEui)
 			if err != nil {
 				if !errors.Is(err, errNoDevice) {
 					return nil, fmt.Errorf("error while searching for device in file: %w", err)
 				}
-
 			}
 			if deviceInfo != nil {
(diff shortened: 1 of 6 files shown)


gateway/gateway.go
```diff
@@ -389,6 +410,22 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 				return nil, err
 			}
 			g.devices[node.NodeName] = mergedNode
+
+			// Check if the device is in the persistent data file, if it is add the OTAA info.
+			deviceInfo, err := g.searchForDeviceInFile(mergedNode.DevEui)
```
JohnN193, line 415: spoke offline about this, technically the DoCommand and handlePacket can both interact with the file at the same time. 

Possibly want a mutex just for reading the file? unsure of the right way to handle, or if its such a small chance of happening maybe we don't worry about it

gateway/gateway.go
```diff
@@ -128,7 +139,16 @@ func NewGateway(
 		started: false,
 	}
 
-	err := g.Reconfigure(ctx, deps, conf)
+	moduleDataDir := os.Getenv("VIAM_MODULE_DATA")
+	filePath := filepath.Join(moduleDataDir, "devicedata.txt")
+	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE, 0o666)
```
JohnN193, line 144: [nit] add a comment for this

gateway/uplink_test.go
```diff
@@ -132,6 +136,7 @@ func TestParseDataUplink(t *testing.T) {
 	_, _, err = g.parseDataUplink(context.Background(), createUnknownDevicePayload())
 	test.That(t, err, test.ShouldNotBeNil)
 	test.That(t, err, test.ShouldBeError, errNoDevice)
+	g.Close(context.Background())
```
JohnN193, line 139: test.That(t, g.Close(context.Background),test.ShouldBeNil) ?

JohnN193 (approved):
LGTM, random nit thoughts I'm having 

- wondering if all of the read/write/create stuff for the deviceInfo file should be in its own devicefile.go file, but the way things are organized now is fine. 

- it was alittle hard to parse the gateway.go DoCommand code, wondering if there could be another helper function in there. Gut says no because of how the `deviceInfo != nil` check is used

gateway/gateway.go
randhid: Nice catch, John - let's get some data. I can imagine Chris running this document with a user that has a file with many sensors; Olivia, can you fold that into your manual tests?

oliviamiller: Added 10 sensors to my config and sometimes got a nil pointer in docommand that locking the mutex fixed. I didn't see any issues with the handlepacket routine writing to the file but it might be because the join requests don't happen very often.

PR #23, opened 2025-01-30
RSDK-9866 readme improvements

oliviamiller:
Based on feedback from esha, explain board and gateway dependencies, made it more clear that the gateway needs to be added in the depends on.


randhid (approved):
@EshaMaharishi let us know if you have any further asks here.

PR #24, opened 2025-02-03
RSDK-9814 Increase packet buffer size to 10

oliviamiller:
When debug was on in the gateway, the line `WARNING: not enough space allocated, fetched 2 packet(s), 1 will be left in RX buffer` showed occasionally. Increased the buffer to 10 to ensure we don't miss any packets, no longer see the warning when manually testing.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unsafe"
 
 	"go.viam.com/rdk/components/board"
 	"go.viam.com/rdk/components/sensor"
@@ -269,6 +270,8 @@ func (g *gateway) captureCOutputToLogs(ctx context.Context, scanner *bufio.Scann
 					g.logger.Error("gateway hardware is misconfigured: unplug the board and wait a few minutes before trying again")
 				case strings.Contains(line, "ERROR"):
 					g.logger.Error(line)
+				case strings.Contains(line, "WARNING"):
+					g.logger.Warn(line)
 				default:
 					g.logger.Debug(line)
 				}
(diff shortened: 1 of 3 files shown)


gateway/gateway.go
```diff
@@ -295,19 +302,23 @@ func (g *gateway) receivePackets(ctx context.Context) {
...
-			g.handlePacket(ctx, payload)
-		default:
+		case -1:
 			g.logger.Errorf("error receiving lora packet")
+		default:
+			// convert from c array to go slice
+			packets := unsafe.Slice((*C.struct_lgw_pkt_rx_s)(unsafe.Pointer(p)), maxRxPkt)
+			for i := range numPackets {
+				// received a LORA packet
+				var payload []byte
```
randhid, line 312: This produces a `[]bytes(nil)`, is this what you want instead of a `[]bytes{}`? both have a length of 0, but one has a nil array.

I think your number of packets check is defensive enough, but just wanted to point this out in case we need to be defensive payload checking and logging for a nile case there as well.

randhid: If it's handled elsewhere just add a comment about where a nil check for the payload is handled, do we have a test for a nil `handlepacket` payload?

gateway/gateway.go
```diff
@@ -295,19 +302,23 @@ func (g *gateway) receivePackets(ctx context.Context) {
...
-			for i := range int(packet.size) {
-				payload = append(payload, byte(packet.payload[i]))
-			}
-			g.handlePacket(ctx, payload)
-		default:
+		case -1:
 			g.logger.Errorf("error receiving lora packet")
+		default:
+			// convert from c array to go slice
+			packets := unsafe.Slice((*C.struct_lgw_pkt_rx_s)(unsafe.Pointer(p)), maxRxPkt)
```
JohnN193, line 309: [supernit] do we need the extra variable? wondering if we can just do this
```suggestion
			packets := unsafe.Slice((*C.struct_lgw_pkt_rx_s)(unsafe.Pointer(p)), int(C.MAX_RX_PKT))
```

gateway/gateway.go
```diff
@@ -295,19 +302,23 @@ func (g *gateway) receivePackets(ctx context.Context) {
...
+			for i := range numPackets {
```
JohnN193, line 310: [q] so before we didn't need this because we set `MAX_RX_PKT = 8`. 

now that `MAX_RX_PKT = 10`, we technically need to loop thru the packets?

JohnN193 (approved):
lgtm, just a question on a code change in the diff

gateway/gateway.go
oliviamiller: yeah I can just do that

gateway/gateway.go
oliviamiller: before we weren't using the MAX_RX_PKT in the C receive function we were doing `return lgw_receive(1, packet);`, so only one packet was received.

randhid (approved):

gateway/gateway.go
randhid: So we'll never have a packet non-zero size with a nil payload?

oliviamiller: the size that is sent in the packet is the payload size, so it should be zero if the payload is nil. So something would have to be wrong with the packet to not match. doesn't hurt to have a nil check just in case, will add.

PR #25, opened 2025-02-13
add dragino sensor

JohnN193:
initial dragino sensor. 

big changes

1. add helpers for making Nodes
2. Allow for gateways to be implicit dependencies(without breaking if made explicit instead)
3. add the dragino sensor with the decoder file included(via http)

also tweaks the gomod/adds some nolints to prevent the linter from breaking everything


draginolht65n/sensor.go
```diff
@@ -0,0 +1,136 @@
...
+		DecoderPath: decoderFilePath,
+		Interval:    conf.Interval,
+		DevEUI:      conf.DevEUI,
+		AppKey:      conf.AppKey,
+		AppSKey:     conf.AppSKey,
+		NwkSKey:     conf.NwkSKey,
+		DevAddr:     conf.DevAddr,
+		Gateways:    conf.Gateways,
+	}
+}
```
JohnN193, line 55: Basically we have other sensors create their own `node.Config`, then we can reuse all of the validation used in the node implementation

draginolht65n/sensor.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	return deps, nil
+}
+
+// LHT65N defines a lorawan node device.
+type LHT65N struct {
+	resource.Named
+	logger logging.Logger
+
+	node        node.Node
```
JohnN193, line 72: the dragino LHT65N wraps a node implementation, so we can easily reuse all of the node functionality

draginolht65n/sensor_test.go
```diff
@@ -0,0 +1,36 @@
+package draginolht65n
```
JohnN193, line 1: pretty much everything is already covered by `node_test.go`. Once we add more dragino specific fields we will probably add more to here.

node/node.go
```diff
@@ -18,26 +16,28 @@ var Model = resource.NewModel("viam", "lorawan", "node")
 
 // Error variables for validation.
 var (
-	errDecoderPathRequired = errors.New("decoder path is required")
```
JohnN193, line 21: exported these so they could be used in other sensors

meta.json
```diff
@@ -16,7 +16,13 @@
...
           "markdown_link": "README.md#configure-the-viamsensornode",
           "short_description": "A sensor compoenent for a LoRaWAN node."
-       }
+       },
+       {
+        "api": "rdk:component:sensor",
+        "model": "viam:lorawan:dragino-LHT65N",
+        "markdown_link": "README.md#configure-the-viamsensornode",
+        "short_description": "A sensor compoenent for a dragino LHT65N LoRaWAN node."
+     }
```
randhid, line 25:
```suggestion
        "short_description": "A sensor component for a dragino LHT65N LoRaWAN node."
     }
```
And the short_description above too.

draginolht65n/LHT65NChirpstack4decoder.js
```diff
@@ -0,0 +1,332 @@
+function decodeUplink(input) {
```
oliviamiller, line 1: Might be better to get it from the github URL in case they make an update? but this is prob fine

node/node_utils.go
```diff
@@ -0,0 +1,157 @@
...
+			expectedFreq,
+			c.ResourceName().AsNamed().Name().Name)
+		return false, nil
+	}
+	return true, nil
+}
+
+// WriteDecoderFile writes an embeded decoderFile into the data folder of the module.
+func WriteDecoderFile(decoderFilename string, decoderFile embed.FS) (string, error) {
+	// Create or open the file used to save device data across restarts.
```
oliviamiller, line 131: delete or edit comment

draginolht65n/sensor.go
```diff
@@ -0,0 +1,136 @@
...
+	ctx context.Context,
+	deps resource.Dependencies,
+	conf resource.Config,
+	logger logging.Logger,
+) (sensor.Sensor, error) {
+	decoderFilePath, err := node.WriteDecoderFile(decoderFilename, decoderFile)
+	if err != nil {
+		return nil, err
+	}
+	n := &LHT65N{
```
oliviamiller, line 86: Instead of wrapping the node object I think you could just create a node.Node here, then you could reuse the node's decoderPath field and the Readings/close functions. Just something to consider not 100% sure it would work.

draginolht65n/sensor.go
```diff
@@ -0,0 +1,136 @@
...
+		Named:  conf.ResourceName().AsNamed(),
+		logger: logger,
+		// This returns the name given to the resource by user
```
oliviamiller, line 89: delete or edit comment

draginolht65n/sensor.go
```diff
@@ -0,0 +1,136 @@
...
+// Reconfigure reconfigure's the node.
+func (n *LHT65N) Reconfigure(ctx context.Context, deps resource.Dependencies, conf resource.Config) error {
+	cfg, err := resource.NativeConfig[*Config](conf)
+	if err != nil {
+		return err
+	}
+
+	nodeCfg := cfg.getNodeConfig(n.decoderPath)
+	if len(deps) == 0 {
+		n.logger.Info("yo no deps")
```
oliviamiller, line 112: remember to change log before merge

draginolht65n/sensor.go
JohnN193: I tried that initially but ran into some issues with reconfigure. I can't reuse node.Node because the reconfigure function is still using a resource.Config that looks for a node.Config with a decoderPath set. Wrapping the object + adding the helpers lets me get around the issue.

draginolht65n/sensor.go
JohnN193: Ah actually meant to delete this one, was using it to debug something

draginolht65n/LHT65NChirpstack4decoder.js
JohnN193: added a decoder from url option! leaving the embed implementation in case someone needs to use it for some reason(no network? idk)

draginolht65n/sensor.go
```diff
@@ -0,0 +1,137 @@
...
+	decoderPath string
+}
+
+func newLHT65N(
+	ctx context.Context,
+	deps resource.Dependencies,
+	conf resource.Config,
+	logger logging.Logger,
+) (sensor.Sensor, error) {
+	httpClient := &http.Client{
```
oliviamiller, line 84: if this is the only time we use the httpClient could we just create it in the WriteDecoderFileFromURL instead of here?

node/node_utils.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	if captureMethodFound && captureFreqHz <= 0 {
+		return 0.0, errors.New("zero or negative capture frequency")
+	}
+	return captureFreqHz, nil
+}
+
+// CheckCaptureFrequency check the resource's capture frequency
+// and reports to the user whether a safe value has been configured.
+func CheckCaptureFrequency(c resource.Config, interval float64, logger logging.Logger) (bool, error) {
```
oliviamiller, line 110: I don't think the bool return value is ever used, could we get rid of it?

gateway/join.go
```diff
@@ -160,8 +160,11 @@ func (g *gateway) generateJoinAccept(ctx context.Context, jr joinRequest, d *nod
 
 	payload := make([]byte, 0)
 	payload = append(payload, 0x20)
+	//nolint:all
 	payload = append(payload, jnLE[:]...)
+	//nolint:all
 	payload = append(payload, netIDLE[:]...)
+	//nolint:all
```
oliviamiller, line 167: tysm for fixing

oliviamiller (approved):
few nits/questions but lgtm!

draginolht65n/sensor.go
JohnN193: I have to create the httpClient outside of the function for the mock in testing to work :(

node/node_utils.go
JohnN193: I can't remember the reason why I added it and I didn't write it down, so yeah lets delete it lol

PR #26, opened 2025-02-14
Bug fix and better logging

oliviamiller:
- Added mutex around data file, John hit an issue where two entries for the same Dev EUI were written to the file. 
- Parsing confirmed data up messages instead of erroring, will add implementation for sending the confirmation downlink in a new PR
- Adding more info to a few of the logs for easier debugging

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -125,6 +125,7 @@ type gateway struct {
 	logReader *os.File
 	logWriter *os.File
 	dataFile  *os.File
+	dataMu    sync.Mutex
 }
 
 // NewGateway creates a new gateway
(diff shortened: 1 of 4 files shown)


gateway/gateway.go
```diff
@@ -458,10 +473,12 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 // searchForDeviceInfoInFile searhces for device address match in the module's data file and returns the device info.
 func (g *gateway) searchForDeviceInFile(devEUI []byte) (*deviceInfo, error) {
 	// read all the saved devices from the file
+	g.dataMu.Lock()
```
randhid, line 476: Why not defer the unlock right after?

randhid: You've got a case where you return and don't unlock in the error.

oliviamiller: thanks for catching, switching to defer

PR #27, opened 2025-02-21
RSDK-10001 Support Downlink messages

JohnN193:
Part 1 to supporting confirmed data up messages, this adds initial support for downlink messages. Users can use do command to provide a frame payload to send to the device. Tested on the em310 and the dragino lht65n.


gateway/downlink.go
```diff
@@ -0,0 +1,212 @@
...
+			return fmt.Errorf("error sending downlink: %w", ctx.Err())
+		}
+		// status of 2 indicates send was successful
+		if int(status) == 2 {
+			break
+		}
+		time.Sleep(2 * time.Millisecond)
+	}
+
+	g.logger.Infof("sent the downlink packet")
```
JohnN193, line 92: we should move this log into `uplink.go` and report which device we sent a downlink to. 

worth keeping at info level since these are not spammy

gateway/downlink.go
```diff
@@ -0,0 +1,212 @@
...
+	// fopts := []byte{
+	// 	0b00111001, // data rate and tx power
+	// 	0xFF,
+	// 	0x00,
+	// 	0x01,
+	// }
+
+	// fopts := []byte{0x3A, 0xFF, 0x00, 0x01}
+
+	// payload = append(payload, fopts...)
```
JohnN193, line 156: [note to self] we wanted to keep these comments for future testing when we support fopts

gateway/downlink.go
```diff
@@ -0,0 +1,212 @@
...
+	err = g.addDeviceInfoToFile(g.dataFile, deviceInfo)
+	if err != nil {
+		return nil, fmt.Errorf("failed to add device info to file: %w", err)
+	}
+
+	return payload, nil
+}
+
+// Helper function to calculate the downlink channel to be used for RX1.
+func findDownLinkChannel(uplinkFreq int) int {
```
JohnN193, line 206: since we return a frequency we should call this `findDownLinkFreq`

gateway/gateway.go
```diff
@@ -592,6 +630,7 @@ func convertToNode(mapNode map[string]interface{}) (*node.Node, error) {
 		return nil, err
 	}
 
+	node.FPort = byte(mapNode["FPort"].(float64))
```
JohnN193, line 633: [nit] `mapNode["FPort"].(int)`? If this doesn't actually matter then feel free to ignore

gateway/join.go
```diff
@@ -204,10 +132,10 @@ func (g *gateway) generateJoinAccept(ctx context.Context, jr joinRequest, d *nod
 
 	// DLSettings byte:
 	// Bit 7: OptNeg (0)
-	// Bits 6-4: RX1DROffset - setting default - offset 0
-	// Bits 3-0: RX2DR - setting default - equal to uplink DR
-	payload = append(payload, 0x00)
-	payload = append(payload, 0x01) // rx delay: 1 second
+	// Bits 6-4: RX1DROffset
+	// Bits 3-0: RX2DR
+	payload = append(payload, 0x08) // Use data rate 8 for rx2 downlinks
```
JohnN193, line 137: does this need to correspond with a value that gets set in `downlink.go`? is there a way for us to make both files use the same variable?

milesightct101/sensor.go
```diff
@@ -74,6 +74,7 @@ func (conf *Config) getNodeConfig() node.Config {
 		NwkSKey:  nwkSKey,
 		DevAddr:  conf.DevAddr,
 		Gateways: conf.Gateways,
+		FPort:    "85",
```
JohnN193, line 77:
```suggestion
		FPort:    "55", // in hex, 85 in decimal
```

node/node.go
```diff
@@ -58,6 +58,7 @@ type Config struct {
 	NwkSKey  string   `json:"network_s_key,omitempty"`
 	DevAddr  string   `json:"dev_addr,omitempty"`
 	Gateways []string `json:"gateways"`
+	FPort    string   `json:"fport,omitempty"`
```
JohnN193, line 61: do we want to update the readme for this? since its a string that we assume to be in hex.

JohnN193 (commented):
LGTM, mostly nits

gateway/gateway.go
oliviamiller: It has to be float64 or it errors

gateway/join.go
oliviamiller: data rate 8 corresponds to SF12 and BW 500k, I can add a comment to explain this

oliviamiller: We could make a structure to convert data rates to sf/bw but dont think its worth doing right now

node/node.go
oliviamiller: yes, adding

gateway/downlink.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, fmt.Errorf("failed to add device info to file: %w", err)
+	}
+
+	return payload, nil
+}
+
+// Helper function to calculate the downlink freq to be used for RX1.
+func findDownLinkFreq(uplinkFreq int) int {
+	// channel number between 0-64
+	upLinkFreqNum := (uplinkFreq - 902300000) / 200000
```
randhid, line 206: Is this done only once or multiple times?

How fast do we need to do this particular function?

Where is this used in this file? Might be best to move it right before its usage.

gateway/gateway.c
```diff
@@ -61,8 +62,15 @@ int setUpGateway(int bus) {
...
-    rfconf.rssi_offset = -215;
+    rfconf.rssi_offset = -215.4;
     rfconf.tx_enable = true;
+    struct lgw_rssi_tcomp_s tcomp;
+    tcomp.coeff_a = 0;
+    tcomp.coeff_b = 0;
+    tcomp.coeff_c = 21.41;
+    tcomp.coeff_d = 2162.56;
+    tcomp.coeff_e = 0;
+    rfconf.rssi_tcomp = tcomp;
```
randhid, line 73: are these frequency or temporal constants?

gateway/gateway.c
```diff
@@ -91,10 +97,49 @@ int setUpGateway(int bus) {
...
+    }
+
+    // the tx gain config contains transmission gain settings for downlinks.
+    // Using the same values as basic station
+    struct lgw_tx_gain_lut_s lut;
+    struct lgw_tx_gain_s txGain[16];
+
+    uint8_t paGain [16] =  {0,0,0,0,0,0,1,1,1,1,1,1,1,1,1};
+    int8_t rf_power [16] = {12, 13,14,15,16,17,18,19,20,21,22,23,24,25,26,27};
+    uint8_t pwr_idx [16] = {15,16,17,19,20,22,1,2,3,4,5,6,7,9,11,14};
```
randhid, line 121: Comments explaining these, please.

gateway/gateway.c
```diff
@@ -91,10 +97,49 @@ int setUpGateway(int bus) {
...
+    for(int i = 0; i<16; i++) {
+        txGain[i].pa_gain = paGain[i];
+        txGain[i].rf_power = rf_power[i];
+        txGain[i].pwr_idx = pwr_idx[i];
+        txGain[i].dig_gain = 0;
+        txGain[i].dac_gain = 3;
+        txGain[i].mix_gain = 5;
+    }
+
+    memcpy(lut.lut, txGain, sizeof(txGain));
```
randhid, line 132: Is there a check that should be performed on lut.lut after the memcpy operation?

gateway/gateway.go
```diff
@@ -338,67 +348,69 @@ func (g *gateway) receivePackets(ctx context.Context) {
...
-					g.handlePacket(ctx, payload)
+					g.handlePacket(ctx, payload, t)
 				}
 			}
 		}
 	}
 }
 
-func (g *gateway) handlePacket(ctx context.Context, payload []byte) {
-	g.receivingWorker.Add(func(ctx context.Context) {
```
randhid, line 360: Nice, I like having fewer workers.

gateway/gateway.go
```diff
@@ -475,22 +487,41 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+			if !ok {
+				return nil, fmt.Errorf("node with name %s not found", name)
+			}
+
+			strPayload, ok := payload.(string)
+			if !ok {
+				return nil, fmt.Errorf("expected a string value but got %v", reflect.TypeOf(strPayload))
+			}
+
+			payloadBytes, err := hex.DecodeString(strPayload)
```
randhid, line 507: Do we have tests that check how this helper behaves with different example payloads that we've seen in our system?

randhid (commented):
Mostly questions and a few nits.

gateway/downlink.go
```diff
@@ -0,0 +1,210 @@
...
+	"go.thethings.network/lorawan-stack/v3/pkg/types"
+	"go.viam.com/utils"
+)
+
+const (
+	joinDelay     = 6         // rx2 delay in seconds for sending join accept message.
+	downlinkDelay = 2         // rx2 delay in seconds for downlink messages.
+	rx2Frequency  = 923300000 // Frequency to send downlinks on rx2 window
+	rx2SF         = 12        // spreading factor for rx2 window
+	rx2Bandwidth  = 0x06      // 500k bandwidth
```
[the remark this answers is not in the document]
oliviamiller: they are defaults for rx2, will add in comment

gateway/join.go
```diff
@@ -334,7 +268,12 @@ func validateMIC(appKey types.AES128Key, payload []byte) error {
 	return nil
 }
 
-func generateKeys(ctx context.Context, devNonce, joinEUI, jn, devEUI, networkID []byte, appKey types.AES128Key) (types.AES128Key, error) {
+type sessionKeys struct {
+	appSKey []byte
+	nwkSKey []byte
+}
```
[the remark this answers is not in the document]
oliviamiller: they need to be the list of bytes in the node struct to go through the docommand so its easier to just make these the same

gateway/uplink.go
```diff
@@ -31,6 +31,27 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte) (strin
...
+
+	// we will send one device downlink from the do command per uplink.
+	if len(device.Downlinks) > 0 {
+		g.logger.Debugf("sending downlink message")
+		payload, err := g.createDownlink(device, device.Downlinks[0])
+		if err != nil {
+			return "", map[string]interface{}{}, fmt.Errorf("failed to create downlink: %w", err)
+		}
+
+		err = g.sendDownlink(ctx, payload, false, packetTime)
```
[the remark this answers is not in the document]
oliviamiller: the join accept downlink has a different RX window than all other downlinks, so the bool is to specify if the downlink is a join accept or not. the comment above is stating that only one downlink is sent per uplink.

node/node_utils.go
```diff
@@ -89,6 +89,14 @@ func (n *Node) ReconfigureWithConfig(ctx context.Context, deps resource.Dependen
 		n.JoinType = "OTAA"
 	}
 
+	if cfg.FPort != "" {
+		val, err := hex.DecodeString(cfg.FPort)
+		if err != nil {
+			return err
+		}
+		n.FPort = val[0]
+	}
```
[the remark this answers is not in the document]
oliviamiller: The fport should only be one byte, I will add a check to validate function to error if its longer than that.

gateway/uplink.go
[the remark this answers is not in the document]
randhid: I see - can we give it a more specific name describing this particular requirement, `isJoinAcceptRxWindow` or something you find suitable.

gateway/gateway.go
oliviamiller: added tests for this

gateway/downlink.go
```diff
@@ -0,0 +1,210 @@
...
+		return nil, err
+	}
+
+	payload = append(payload, encrypted...)
+
+	mic, err := crypto.ComputeLegacyDownlinkMIC(
+		types.AES128Key(device.NwkSKey), *types.MustDevAddr(device.Addr), device.FCntDown+1, payload)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
oliviamiller: since we are supporting lorawan version 1.0.3 we use the legacy functions, the mic and keys calculation are different in lorawan 1.1 and use a different function

gateway/downlink.go
oliviamiller: we didnt actually end up using this because we used RX2 window, but wanted to keep here in case we ever need to use RX1 in the future. ill add a comment.

gateway/gateway.c
oliviamiller: temperature, its a polynomial equation to adjust for temperature variation, we got the values from basic station. I don't think this is actually needed right now for our setup but we added everything basic station had to the C setup.

gateway/gateway.c
```diff
@@ -91,10 +97,49 @@ int setUpGateway(int bus) {
...
+
+    if(lgw_txgain_setconf(0, &lut) != LGW_HAL_SUCCESS) {
+        return 6;
+    }
+
     // start the gateway.
     int res = lgw_start();
     if (res != LGW_HAL_SUCCESS) {
-        return 5;
+        return 7;
```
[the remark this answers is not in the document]
oliviamiller: yes, since we added more steps above to the configuration I added more error codes

gateway/gateway.c
oliviamiller: not sure I know what you mean, we use it below in lgw_txgain_setconf

gateway/gateway.go
```diff
@@ -315,8 +323,10 @@ func (g *gateway) receivePackets(ctx context.Context) {
 			return
 		default:
 		}
+		g.mu.Lock()
 		numPackets := int(C.receive(p))
-
+		g.mu.Unlock()
+		t := time.Now()
```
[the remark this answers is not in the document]
oliviamiller: we were locking on send and receive to avoid doing them at the same time but we ended up taking out the lock on send, think we can take it out here too

gateway/downlink.go
```diff
@@ -0,0 +1,210 @@
...
+	for {
+		C.lgw_status(txPkt.rf_chain, 1, &status)
+		if err := ctx.Err(); err != nil {
+			return fmt.Errorf("error sending downlink: %w", ctx.Err())
+		}
+		// status of 2 indicates send was successful
+		if int(status) == 2 {
+			break
+		}
+		time.Sleep(2 * time.Millisecond)
```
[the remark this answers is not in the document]
oliviamiller: the lgw_send command has 3-10ms latency

gateway/downlink.go
randhid: When we eventually lint I think `revive` will complain about the unused function.

gateway/gateway.c
randhid: I see, I just wanted to make sure there was a failure case that we could capture, seems like there is the appropriate place to do so.

gateway/downlink.go
```diff
@@ -0,0 +1,210 @@
...
+	txPkt.payload = cPayload
+
+	var waitDuration time.Duration
+	switch isJoinAccept {
+	case true:
+		// 47709/32*time.Microsecond is the internal delay of sending a packet
+		waitDuration = (joinDelay * time.Second) - (time.Since(packetTime)) - 47709/32*time.Microsecond
+	default:
+		waitDuration = (downlinkDelay * time.Second) - (time.Since(packetTime)) - 47709/32*time.Microsecond
+	}
```
randhid, line 68: If else with a bool is more idiomatic.

```suggestion
        // 47709/32*time.Microsecond is the internal delay of sending a packet
        waitDuration := (downlinkDelay * time.Second) - (time.Since(packetTime)) - 47709/32*time.Microsecond
	if isJoinAccept {
		waitDuration = (joinDelay * time.Second) - (time.Since(packetTime)) - 47709/32*time.Microsecond
	}
```

gateway/downlink.go
oliviamiller: the linter is passing so i dont think we check for that, should we be?

gateway/uplink.go
```diff
@@ -31,6 +31,27 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte) (strin
...
+		if err != nil {
+			return "", map[string]interface{}{}, fmt.Errorf("failed to send downlink: %w", err)
+		}
+
+		g.logger.Infof("sent the downlink packet to %s", device.NodeName)
```
randhid, line 49: Is this noisy?

node/node.go
```diff
@@ -160,6 +167,11 @@ type Node struct {
 	NodeName    string
 	gateway     sensor.Sensor
 	JoinType    string
+
+	FCntDown uint32
+	FPort    byte // for downlinks, only required when frame payload exists.
+
```
randhid, line 173:
```suggestion
```

randhid (approved):
LGTM!

gateway/uplink.go
JohnN193: this could be noisy once we support confirmed data, but we can maybe wait until that pr to evaluate

PR #28, opened 2025-02-25
[RSDK-10087] add ct101 and em310 sensor model

JohnN193:
https://viam.atlassian.net/browse/RSDK-10087
https://viam.atlassian.net/browse/RSDK-10088

Will update dragino sensors to match once the decoder issue is resolved. There may be further optimizations for setting up specific sensors(i.e. move around more shared code) we can find once we have downlinks working

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,51 @@ The node registers itself with the gateway so the gateway will recognize message
 | app_s_key | string | yes | Application Session Key (16 bytes in hex) Used to decrypt uplink messages. Default can normally be found on the node's datasheet or box. |
 | network_s_key | string | yes | Network Session Key (16 bytes in hex) Used to decypt uplink messages. Default can normally be found on the node's datasheet or box. |
 
+## Configure the `viam:lorawan:Milesight-CT101`
+
+Example OTAA node configuration:
+
+```json
+{
+  "join_type": "OTAA",
+  "dev_eui": "0123456789ABCDEF"
+}
+```
+
+Example ABP node configuration:
+
+```json
+{
+  "join_type": "ABP",
+  "dev_addr": "01234567"
+}
+```
+
+### Common Attributes
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| join_type | string | no | Join type ("OTAA" or "ABP"). Defaults to "OTAA" |
+| uplink_interval_mins | float64 | no | Expected interval between uplink messages sent by the node. The default is **10** minutes.
+
+The gateway component must be added as a dependency in the `Depends on` drop down.
+The node registers itself with the gateway so the gateway will recognize messages from the node.
+
+### OTAA Attributes
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| dev_eui | string | yes | Device EUI (8 bytes in hex). Unique indentifer for the node. Can be found printed on your device or on the box.|
+| app_key | string | no | Application Key (16 bytes in hex). Used to securely join the network. The default is **5572404C696E6B4C6F52613230313823**. |
+
+### ABP Attributes
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| dev_addr | string | yes | Device Address (4 bytes in hex). Used to identify uplink messages. Can normally be found on datasheet or box. |
+| app_s_key | string | yes | Application Session Key (16 bytes in hex) Used to decrypt uplink messages. Default is **5572404C696E6B4C6F52613230313823**. |
+| network_s_key | string | yes | Network Session Key (16 bytes in hex) Used to decypt uplink messages. Default is **5572404C696E6B4C6F52613230313823**. |
+
 ## Configure the `viam:lorawan:Milesight-EM310-tilt`
 
 Example OTAA node configuration:
--- a/README.md
+++ b/README.md
@@ -110,6 +110,51 @@ The node registers itself with the gateway so the gateway will recognize message
 | app_s_key | string | yes | Application Session Key (16 bytes in hex) Used to decrypt uplink messages. Default can normally be found on the node's datasheet or box. |
 | network_s_key | string | yes | Network Session Key (16 bytes in hex) Used to decypt uplink messages. Default can normally be found on the node's datasheet or box. |
 
+## Configure the `viam:lorawan:Milesight-EM310-tilt`
+
+Example OTAA node configuration:
+
+```json
+{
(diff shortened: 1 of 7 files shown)


oliviamiller (approved):

milesightct101/sensor.go
```diff
@@ -0,0 +1,160 @@
...
+	// ABP defaults.
+	defaultNwkSKey = "5572404C696E6B4C6F52613230313823"
+	defaultAppSKey = "5572404C696E6B4C6F52613230313823"
+)
+
+// defaultIntervalMin is how often the CT101 will send an uplink.
+var defaultIntervalMin = 10. // minutes
+
+// Model represents a Milesight-CT101 lorawan node model.
+var Model = node.LorawanFamily.WithModel("milesight-ct101")
```
randhid, line 29: we should move the families to a directory level `lorawan` package

milesightct101/sensor.go
```diff
@@ -0,0 +1,160 @@
...
+		AppSKey:     appSKey,
+		NwkSKey:     nwkSKey,
+		DevAddr:     conf.DevAddr,
+		Gateways:    conf.Gateways,
+	}
+}
+
+// Validate ensures all parts of the config are valid.
+func (conf *Config) Validate(path string) ([]string, error) {
+	nodeConf := conf.getNodeConfig("fixed")
```
randhid, line 84: could this error?

milesightct101/sensor.go
```diff
@@ -0,0 +1,160 @@
...
+		return err
+	}
+
+	err = node.CheckCaptureFrequency(conf, *cfg.Interval, n.logger)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
randhid, line 150: if these are common to nodes - also move to higher level `lorawan` package

milesightem310/em310.go
```diff
@@ -0,0 +1,159 @@
+// package milesightem310 implements em310 node
```
randhid, line 1:
```suggestion
// Package milesightem310 implements em310 node
```

randhid (commented):
If the node code is similar enough - I think we should think about having common functions at a top level `lorawan` package at this point and just having the inits in a `nodes` package + modular main code in a `cmd/module`  folder. Okay to do in a separate ticket.

That is now the preferred style for go modules - the go generator style has changed.

README.md
```diff
@@ -110,6 +112,101 @@ The node registers itself with the gateway so the gateway will recognize message
...
+  "gateways": ["gateway-1"]
+}
+```
+
+### Common Attributes
+
+| Name | Type | Required | Description |
+|------|------|----------|-------------|
+| join_type | string | no | Join type ("OTAA" or "ABP"). Defaults to "OTAA" |
+| uplink_interval_mins | float64 | no | Expected interval between uplink messages sent by the node. The default is **10** minutes. |
```
randhid, line 142: let's just have one common configure section for the sensors

milesightct101/sensor_test.go
```diff
@@ -0,0 +1,34 @@
...
+		JoinType: node.JoinTypeOTAA,
+		DevEUI:   testDevEUI,
+		AppKey:   testAppKey,
+		Gateways: []string{testGatewayName},
+	}
+	deps, err := conf.Validate("")
+	test.That(t, err, test.ShouldBeNil)
+	test.That(t, len(deps), test.ShouldEqual, 1)
+	test.That(t, deps[0], test.ShouldEqual, testGatewayName)
+}
```
randhid, line 34: Let's just move this to one test in a higher level package

JohnN193:
> Top level Lorawan package

yeah we talked about doing this + a shared config for specific sensors for for a lil bit yesterday. The main concerns were:

1. we still don’t quite know what downlinks will need to look like, as those should be significantly different from sensor to sensor
2. Users configuring their own sensor may need to set some other attributes that we don’t need, which could make it harder for them to create their own sensors. Not a huge deal tbh 
3. The majority of the code already runs through the generic `node`, so having multiple top level packages might be confusing 


There’s def ways to remove more of the shared code, but I think it’s worth waiting until after downlinks get solved so we can plan around those

randhid:
> > Top level Lorawan package
> 
> yeah we talked about doing this + a shared config for specific sensors for for a lil bit yesterday. The main concerns were:
> 
>     1. we still don’t quite know what downlinks will need to look like, as those should be significantly different from sensor to sensor
> 
>     2. Users configuring their own sensor may need to set some other attributes that we don’t need, which could make it harder for them to create their own sensors. Not a huge deal tbh
> 
>     3. The majority of the code already runs through the generic `node`, so having multiple top level packages might be confusing
> 
> 
> There’s def ways to remove more of the shared code, but I think it’s worth waiting until after downlinks get solved so we can plan around those

that's okay with me - please make a ticket to track that in the epic

README.md
randhid: @michaellee1019 do you know any markdown magic that could help us here?

Basically we have a bunch of sensors that have identical config attributes, but the value of the keys for each sensor family is different.

oliviamiller:
> > > Top level Lorawan package
> > 
> > 
> > yeah we talked about doing this + a shared config for specific sensors for for a lil bit yesterday. The main concerns were:
> > ```
> > 1. we still don’t quite know what downlinks will need to look like, as those should be significantly different from sensor to sensor
> > 
> > 2. Users configuring their own sensor may need to set some other attributes that we don’t need, which could make it harder for them to create their own sensors. Not a huge deal tbh
> > 
> > 3. The majority of the code already runs through the generic `node`, so having multiple top level packages might be confusing
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There’s def ways to remove more of the shared code, but I think it’s worth waiting until after downlinks get solved so we can plan around those
> 
> that's okay with me - please make a ticket to track that in the epic

ticket made https://viam.atlassian.net/browse/RSDK-10103

README.md
JessamyT: Hmm. Since there are only 3 common attributes, and since inline docs are a thing now, maybe it'd be best to copy these three lines into the table for each of the models so that the inline docs can just show one table for each. I don't love the duplicated content for people looking at the whole readme and for maintenance purposes, but definitely feels like that'd make a better inline experience.

randhid: Okay - we'll figure out if there's magic markdown to make this better another day!

randhid (approved):

PR #29, opened 2025-02-26
RSDK-10107 Use accurate sleep and ignore duplicate packets

oliviamiller:
Because there is only a 20 microsecond allowance for the rx delays, we should use accurate sleep to ensure the join accept makes it in time for the device's RX window. Additionally, filtering out duplicate packets we receive.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -8,6 +8,7 @@ package gateway
 #include "../sx1302/libloragw/inc/loragw_hal.h"
 #include "gateway.h"
 #include <stdlib.h>
+#include <string.h>
 */
 import "C"
 
@@ -18,6 +19,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -334,6 +336,14 @@ func (g *gateway) receivePackets(ctx context.Context) {
 				if packets[i].size == 0 {
 					continue
 				}
+
+				// don't process duplicates
+				if numPackets > 1 && i > 0 {
+					if isSamePacket(packets[i-1], packets[i]) {
+						g.logger.Debugf("skipped duplicate packet")
+						continue
+					}
+				}
 				// Convert packet to go byte array
 				for j := range int(packets[i].size) {
 					payload = append(payload, byte(packets[i].payload[j]))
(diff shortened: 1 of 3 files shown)


gateway/join.go
```diff
@@ -81,9 +81,8 @@ func (g *gateway) handleJoin(ctx context.Context, payload []byte) error {
 	txPkt.payload = cPayload
 
 	// send on rx2 window - opens 6 seconds after join request.
-	if !utils.SelectContextOrWait(ctx, time.Second*joinRx2WindowSec) {
-		return nil
-	}
+	waitDuration := (joinRx2WindowSec * time.Second) - (time.Since(t))
+	accurateSleep(ctx, waitDuration)
```
JohnN193, line 85: [nit] have a comment that we will move this to the generic downlink function when that is ready

gateway/join.go
```diff
@@ -96,6 +95,36 @@ func (g *gateway) handleJoin(ctx context.Context, payload []byte) error {
 	return nil
 }
 
+// According to lorawan docs, the rx windows have an error range of 20 microseconds, so
+// we need to sleep for a more accurate amount of time.
+func accurateSleep(ctx context.Context, duration time.Duration) bool {
```
JohnN193, line 100: thank you @penguinland

JohnN193 (approved):

gateway/gateway.go
```diff
@@ -469,6 +479,23 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+//	-- freq should be same
+//	-- datarate should be same
+//	-- payload should be same
+func isSamePacket(p1, p2 C.struct_lgw_pkt_rx_s) bool {
+	if math.Abs(float64(p1.count_us-p2.count_us)) <= 24 &&
+		p1.freq_hz == p2.freq_hz &&
+		p1.datarate == p2.datarate &&
+		//nolint:gocritic
+		C.memcmp(unsafe.Pointer(&p1.payload[0]), unsafe.Pointer(&p2.payload[0]), C.size_t(len(p1.payload))) == 0 {
+		return true
```
randhid, line 494: We should isolate C calls more stringently from the go code, pleas make a ticket fro cleanup/test coverage of Cgo code.

randhid (approved):
Nice! We should - when we refactor this - go over where we're using C code and how to best isolate it from go.  Please add that to the ticket for refactoring.

oliviamiller:
> Nice! We should - when we refactor this - go over where we're using C code and how to best isolate it from go. Please add that to the ticket for refactoring.

sounds good, made ticket https://viam.atlassian.net/browse/RSDK-10108

PR #31, opened 2025-03-06
RSDK-10089 allow decoder_path attribute to be link

oliviamiller:
Allows the decoder_path in generic node model to be a HTTP link. Tested that all 4 node models still behave as expected.

### DIFF
--- a/draginolht65n/sensor.go
+++ b/draginolht65n/sensor.go
@@ -4,8 +4,6 @@ package draginolht65n
 import (
 	"context"
 	"embed"
-	"os"
-	"path/filepath"
 
 	"github.com/viam-modules/gateway/node"
 	"go.viam.com/rdk/components/sensor"
@@ -44,14 +42,11 @@ func init() {
 		})
 }
 
-func (conf *Config) getNodeConfig() node.Config {
-	moduleDataDir := os.Getenv("VIAM_MODULE_DATA")
-	filePath := filepath.Join(moduleDataDir, decoderFilename)
-
+func (conf *Config) getNodeConfig(decoderFilePath string) node.Config {
 	return node.Config{
 		JoinType: conf.JoinType,
 		Interval: conf.Interval,
-		Decoder:  filePath,
+		Decoder:  decoderFilePath,
 		DevEUI:   conf.DevEUI,
 		AppKey:   conf.AppKey,
 		AppSKey:  conf.AppSKey,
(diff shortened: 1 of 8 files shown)


milesightem310/em310.go
```diff
@@ -13,8 +11,7 @@ import (
 )
 
 const (
-	decoderFilename = "LHT65NChirpstack4decoder.js"
-	decoderURL      = "https://raw.githubusercontent.com/Milesight-IoT/SensorDecoders/refs/heads/main/EM_Series/" +
+	decoderURL = "https://raw.githubusercontent.com/Milesight-IoT/SensorDecoders/refs/heads/main/EM_Series/" +
```
randhid, line 14: Is this MileSights? do we ever expect this to disappear?

randhid (approved):

milesightem310/em310.go
oliviamiller: yeah its possible they could delete or move it, but wanted to use the link because they also make updates to them.

randhid: If there's a way to pin to a specific version great - or a test that can just exist if this URL exists with each pr.

randhid: Would suck to find out at run time that they changed the URL completely.

draginolht65n/sensor.go
```diff
@@ -83,16 +84,15 @@ func newLHT65N(
 	conf resource.Config,
 	logger logging.Logger,
 ) (sensor.Sensor, error) {
-	decoderFilePath, err := node.WriteDecoderFile(decoderFilename, decoderFile)
+	_, err := node.WriteDecoderFile(decoderFilename, decoderFile)
```
JohnN193, line 87: we should still use `decoderFilePath`, in case we have to change the how the path is written in `WriteDecoderFile`.

it makes `getNodeConfig` need the input still, but I think its fine if the dragino has slightly different functions for now.

node/node_utils.go
```diff
@@ -58,7 +61,26 @@ func (n *Node) ReconfigureWithConfig(ctx context.Context, deps resource.Dependen
 		n.AppSKey = appSKey
 	}
 
-	n.DecoderPath = cfg.DecoderPath
+	n.DecoderPath = cfg.Decoder
+	// if the decoder path is a url, save the file
+	if isValidURL(n.DecoderPath) {
```
JohnN193, line 66: add a check to ensure the DecoderPath is a file if we were not given a url

milesightem310/em310.go
JohnN193: if it is possible then cool, otherwise we can probably store some backups on the google drive or somewhere.

oliviamiller: I changed the links to use the commit hash, we just may have to update them in the future if they make updates to the decoders

node/node_utils.go
JohnN193: you could actually check that the file is a js file too

PR #34, opened 2025-03-12
[RSDK-10241] Support setting interval downlinks for sensors

JohnN193:
https://viam.atlassian.net/browse/RSDK-10241

This PR also adds significantly more tests to the specific nodes. 90% of which are effectively the same but some names/parts switched around

the important change is adding the `DoCommand` to each of the nodes and the `SendDownlink` method on the generic node


node/node.go
```diff
@@ -43,9 +44,14 @@ const (
...
 	JoinTypeABP = "ABP"
+	// DownlinkKey is the DoCommand Key to send a payload to the gateway.
+	DownlinkKey = "send_downlink"
+	// GatewaySendDownlinkKey is the DoCommand Key for the gateway model to queue a downlink.
+	GatewaySendDownlinkKey = "add_downlink_to_queue"
 )
 
-var noReadings = map[string]interface{}{"": "no readings available yet"}
+// NoReadings is the return for a sensor that has not received data.
+var NoReadings = map[string]interface{}{"": "no readings available yet"}
```
JohnN193, line 54: exported this to use in sensor specific tests.

milesightct101/sensor_test.go
```diff
@@ -16,7 +25,255 @@ const (
...
+		}
+		return map[string]interface{}{}, nil
+	}
+	mockGateway.ReadingsFunc = func(ctx context.Context, cmd map[string]interface{}) (map[string]interface{}, error) {
+		readings := make(map[string]interface{})
+		readings["test-ct101"] = testNodeReadings
+		return readings, nil
+	}
+	return mockGateway
+}
```
JohnN193, line 50: using this in every sensor. Didn't seem worth to figure out the right way to make the helper, and making a test/helper thats too complicated isn't helpful anymore

README.md
```diff
@@ -117,6 +119,20 @@ The node registers itself with the gateway so the gateway will recognize message
...
+#### Send a downlink
+
+This command will send a generic downlink payload to the gateway. the string is expected to be a set of bytes in hex.
+
+```json
+{
+  "send_downlink": "<BYTESINHEX>"
+}
+```
+
```
JohnN193, line 135: I only added documentation for the nodes. I chose not to add documentation for the gateway downlink DoCommand, but we can add it if people want.

node/node.go
```diff
@@ -44,9 +45,14 @@ const (
 	JoinTypeOTAA = "OTAA"
 	// JoinTypeABP is the ABP Join type.
 	JoinTypeABP = "ABP"
+	// DownlinkKey is the DoCommand Key to send a payload to the gateway.
+	DownlinkKey = "send_downlink"
+	// GatewaySendDownlinkKey is the DoCommand Key for the gateway model to queue a downlink.
+	GatewaySendDownlinkKey = "add_downlink_to_queue"
```
JohnN193, line 51: @oliviamiller instead of using `senddown` on the gateway, I changed the key to be this. can easily switch it back if we prefer the smaller message, but I kinda prefer the more verbose response to the user

draginolht65n/sensor.go
```diff
@@ -121,6 +125,15 @@ func (n *LHT65N) Reconfigure(ctx context.Context, deps resource.Dependencies, co
...
 	}
 
+	// set the interval if one was provided
+	// we do not send a default in case the user has already set an interval they prefer
+	if cfg.Interval != nil && *cfg.Interval != 0 {
+		_, err = n.addIntervalToQueue(ctx, *nodeCfg.Interval, false)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
oliviamiller: agree with this behavior, we don't want to override something the user has already configured outside of viam

draginolht65n/sensor.go
```diff
@@ -139,3 +151,30 @@ func (n *LHT65N) Close(ctx context.Context) error {
...
+
+// DoCommand implements the DoCommand for the LHT65N.
+func (n *LHT65N) DoCommand(ctx context.Context, cmd map[string]interface{}) (map[string]interface{}, error) {
+	testOnly := node.CheckTestKey(cmd)
+
+	if interval, intervalSet := cmd[intervalKey]; intervalSet {
+		if intervalFloat, floatOk := interval.(float64); floatOk {
+			return n.addIntervalToQueue(ctx, intervalFloat, testOnly)
+		}
+		return map[string]interface{}{}, fmt.Errorf("error parsing payload, expected float got %v", interval)
```
oliviamiller, line 163: I think you need to do reflect.TypeOf(interval) to get what you want here

draginolht65n/sensor_test.go
```diff
@@ -16,7 +24,234 @@ const (
...
+	deps[encoder.Named(testGatewayName)] = mockGateway
+
+	tmpDir := t.TempDir()
+	testDecoderPath := fmt.Sprintf("%s/%s", tmpDir, decoderFilename)
+	t.Setenv("VIAM_MODULE_DATA", tmpDir)
+
+	// Create the file, as http tests were already tested in node_test.go
+	file, err := os.Create(testDecoderPath)
+	test.That(t, err, test.ShouldBeNil)
+	defer file.Close()
```
oliviamiller, line 173: could make a helper for these lines since its repeated in every test

draginolht65n/sensor_test.go
```diff
@@ -16,7 +24,234 @@ const (
...
+
+		// we should not receive a node success message
+		nodeResp, nodeOk := resp[node.DownlinkKey].(map[string]interface{})
+		test.That(t, nodeOk, test.ShouldBeFalse)
+		test.That(t, nodeResp, test.ShouldBeNil)
+	})
+	t.Run("Test successful interval downlink DoCommand to Gateway", func(t *testing.T) {
+		// testKey controls whether we send bytes to the gateway. used for debugging.
+		// req := map[string]interface{}{testKey: "", DownlinkKey: "bytes"}
+		req := map[string]interface{}{intervalKey: 10.0}
```
oliviamiller, line 209: is the testKey not supposed to be used in every test?

JohnN193: no, testKey is meant to let people see what bytes/messages are getting sent without adding them to the queue.

oliviamiller (approved):

gateway/gateway.go
```diff
@@ -533,7 +531,7 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 			g.readingsMu.Unlock()
 		}
 	}
-	if payload, ok := cmd[sendDownlinkKey]; ok {
+	if payload, ok := cmd[node.GatewaySendDownlinkKey]; ok {
```
randhid, line 534: I don't love that the node has a SendDownlink variable - but I think we should just have a top level lorawan package containing all the functionality of node and gateway when you refactor this module.

milesightct101/sensor.go
```diff
@@ -129,6 +136,15 @@ func (n *CT101) Reconfigure(ctx context.Context, deps resource.Dependencies, con
...
 	}
 
+	// set the interval if one was provided
+	// we do not send a default in case the user has already set an interval they prefer
+	if cfg.Interval != nil && *cfg.Interval != 0 {
+		_, err = n.addIntervalToQueue(ctx, *nodeCfg.Interval, false)
+		if err != nil {
+			return err
+		}
+	}
```
randhid, line 146: This is repeated, in a few places, please make it a function on cfg if that is a common config.

milesightem310/sensor.go
```diff
@@ -144,3 +160,51 @@ func (n *em310Tilt) Close(ctx context.Context) error {
...
+	for _, b := range bs {
+		intervalString = fmt.Sprintf("%s%02x", intervalString, b)
+	}
+
+	if testOnly {
+		return map[string]interface{}{intervalKey: intervalString}, nil
+	}
+
+	return n.node.SendDownlink(ctx, intervalString, false)
+}
```
randhid, line 200: Is this also repeated code?

node/node.go
```diff
@@ -283,9 +290,24 @@ func (n *Node) Readings(ctx context.Context, extra map[string]interface{}) (map[
...
+	testOnly := CheckTestKey(cmd)
+
+	if payload, payloadSet := cmd[DownlinkKey]; payloadSet {
+		if payloadString, payloadOk := payload.(string); payloadOk {
+			return n.SendDownlink(ctx, payloadString, testOnly)
+		}
+		return map[string]interface{}{}, fmt.Errorf("error parsing payload, expected string got %v", reflect.TypeOf(payload))
+	}
+
+	return resp, nil
```
randhid, line 312: Was there an observed race with Reconfigure? Does `n.SendDownlink` require a lock?
If we're joining and someone commands a SendDownlink - is the module able to handle that case? I assume the LoRaWan network won't care.

randhid (commented):
Can you combine/DRY up some of the repeated code?

milesightct101/sensor.go
JohnN193: `cfg` is specific to every model and `addIntervalToQueue` is implemented differently for every model. I could maybe to simplify this further by passing in `addIntervalToQueue` to a different function

milesightem310/sensor.go
JohnN193: nope, the implemetation for `addIntervalToQueue` is different for every model. the repeated parts are 

```
if testOnly {
		return map[string]interface{}{intervalKey: intervalString}, nil
	}

	return n.node.SendDownlink(ctx, intervalString, false)
```

node/node.go
JohnN193: so I am not aware about reconfigure, but when a downlink is sent the following behaviors can happen. 

1. if the module just started and we have not [registered the node](https://github.com/viam-modules/lorawan/blob/fddf22157ce9e60f2877e8dbc05f3287d3ea6c7c/gateway/gateway.go#L492) the device yet(or loaded from the file) then the downlink will [fail to add](https://github.com/viam-modules/lorawan/blob/fddf22157ce9e60f2877e8dbc05f3287d3ea6c7c/gateway/gateway.go#L543-L546) 
2. if the node was registered but the device has not sent a join request yet, the downlink will be [added to the queue](https://github.com/viam-modules/lorawan/blob/fddf22157ce9e60f2877e8dbc05f3287d3ea6c7c/gateway/gateway.go#L558) and sent when we receive the [first uplink from the device](https://github.com/viam-modules/lorawan/blob/fddf22157ce9e60f2877e8dbc05f3287d3ea6c7c/gateway/uplink.go#L64)
3. if the sensor has already joined, then the downlink is added to the queue and sent on next uplink.

the gateway will lock when receiving DoCommands, so we shouldn't have to worry about multiple downlinks getting registered at the same time.

For my own info, can a DoCommand/request be sent while a component is in a reconfigure state? I kinda assumed no based on the behavior of the frontend but I haven't actually tested it.

milesightct101/sensor.go
randhid: you could can pass the pointer in?

node/node.go
randhid: Yes a DoCommand and Reconfigure can be parallel if the resource has already built and is not an always rebuild resource. That's usually the trickiest deadlock/race to detect for me personally so I ask this question whenever I remember it 😆

gateway/gateway.go
JohnN193: yeah I needed to have a key that both `gateway` and `node` can read, but gateway already imports node. once we get to refactoring we can maybe use an eng focus to map out the package structure and what we want implementers to have to implement.

milesightct101/sensor.go
JohnN193: we spoke offline, at a future date we can try moving `addIntervalToQueue` to `node`, that takes a struct that users fill in to handle endianness/units/size etc.

JohnN193: we can also try using generics to remove the configs

node/node.go
JohnN193: thinking more on this, a user technically can get into trouble if the gateway/dependencies were to change during a reconfigure. the nodes could send downlinks to the wrong gateway. 

I can try putting in a lock in the `node` struct to prevent multiple sends & the reconfigure issue.

randhid: Nice find John! Up to you whether this is done here or in a follow up pr - happy with either, just please make a ticket if you prefer merging this in and addressing the issue of sending a downlink (? I would have expected uplink) to a wrong gateway in a separate pr.

randhid (approved):

PR #35, opened 2025-03-13
RSDK-10001 RSDK-9536 support confirmed data uplinks

oliviamiller:
if the device sends a confirmed data uplink packet, the gateway must send a downlink with an ACK sent to confirm it received the data. Tested with em310-tilt, it resends the data if the data is not confirmed, and the data is no longer present with this change.


gateway/uplink.go
```diff
@@ -33,23 +33,32 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
-		g.logger.Debugf("sending downlink message to %s", device.NodeName)
-		payload, err := g.createDownlink(device, device.Downlinks[0])
+		// we will send one device downlink from the do command per uplink.
+		downlinkFramePayload = device.Downlinks[0]
+		// remove the downlink we are about to send from the queue.
+		device.Downlinks = device.Downlinks[1:]
+	}
+
+	if downlinkFramePayload != nil || sendAck {
+		downlinkPayload, err := g.createDownlink(device, downlinkFramePayload, sendAck)
```
JohnN193, line 51: [note to self] we talked about this function getting alittle bit messy as we look to handle different kinds of downlinks that don't get queued(i.e. ack, device time, etc) 

no idea what we should do but wanted to write this down

gateway/downlink.go
```diff
@@ -150,22 +154,24 @@ func (g *gateway) createDownlink(device *node.Node, framePayload []byte) ([]byte
...
 	// 30 seconds
 	// framePayload := []byte{0x01, 0x00, 0x00, 0x1E} //  dragino
 	// framePayload := []byte{0xff, 0x10, 0xff} //tilt reset
 
-	encrypted, err := crypto.EncryptDownlink(
-		types.AES128Key(device.AppSKey), *types.MustDevAddr(device.Addr), device.FCntDown+1, framePayload)
-	if err != nil {
-		return nil, err
+	if framePayload != nil {
+		if device.FPort == 0 {
```
JohnN193, line 163: we should also check the FPort is non zero in `register_device do command`

gateway/uplink.go
```diff
@@ -33,23 +33,32 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
 		if err != nil {
 			return "", map[string]interface{}{}, fmt.Errorf("failed to create downlink: %w", err)
 		}
-
-		err = g.sendDownlink(ctx, payload, false, packetTime)
+		err = g.sendDownlink(ctx, downlinkPayload, false, packetTime)
```
randhid, line 55:
```suggestion
		if err = g.sendDownlink(ctx, downlinkPayload, false, packetTime); err!= nil {...}
```

gateway/uplink.go
```diff
@@ -33,23 +33,32 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 
 	g.logger.Debugf("received uplink from %s", device.NodeName)
 
-	// we will send one device downlink from the do command per uplink.
+	sendAck := false
+	// confirmed data up, send ACK bit in downlink
+	if phyPayload[0] == 0x80 {
```
JohnN193, line 38: as a part of the refactor we should try replacing all of these (`0x80`) with constants. Would hopefully make the code easier to read and more obvious what we are checking for in different spots

JohnN193 (approved):
LGTM, mostly comments/discussion on things we can try planning for later

gateway/uplink.go
```diff
@@ -33,23 +33,32 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
+		sendAck = true
+	}
```
randhid, line 40: does this work?

```suggestion
	// confirmed data up, send ACK bit in downlink
	sendAck := phyPayload[0] == 0x80 {
```

gateway/downlink.go
```diff
@@ -131,14 +131,18 @@ func (g *gateway) createDownlink(device *node.Node, framePayload []byte) ([]byte
 
 	payload = append(payload, devAddrLE...)
 
-	// FCtrl: ADR (0), RFU (0), ACK (0), FPending (0), FOptsLen (0000)
-	payload = append(payload, 0x00)
+	//  FCtrl: ADR (0), RFU (0), ACK(0/1), FPending (0), FOptsLen (0000)
```
randhid, line 134: comment explaining why we append this first. just in case someone is moving code around in the future.

oliviamiller: not sure I know what you mean, a comment with the order that everything must be appended is at the top the function

randhid: Literally right in front of my face - thanks!

JohnN193: maybe link to the documentation and mention what section this is in?

gateway/downlink.go
oliviamiller: we don't validate any of the other fields in the do command so it feels weird to me to do it there, I could add another check to the validate function to error if the user puts zero

gateway/uplink.go
randhid: Whatever linters you guys turn on will have a field day.

gateway/downlink.go
JohnN193: that might be fine, big thing I want to avoid is a user suddenly seeing errors because their custom node is misconfigured somehow. super unlikely but would be better to surface the error sooner rather than at downlink time.

gateway/uplink.go
JohnN193: eh rdk turns off magic number linter too so I don't see us using that one

gateway/uplink.go
oliviamiller: yes!

node/node.go
```diff
@@ -93,8 +94,14 @@ func (conf *Config) Validate(path string) ([]string, error) {
...
 
-	if len(conf.FPort) > 2 {
-		return nil, resource.NewConfigValidationError(path, ErrFPortTooLong)
+	if conf.FPort != "" {
+		fPort, err := hex.DecodeString(conf.FPort)
+		if err != nil {
+			return nil, resource.NewConfigValidationError(path, ErrInvalidFPort)
+		}
+		if fPort[0] == byte(0x00) || fPort[0] > byte(0xDF) || len(fPort) > 1 {
+			return nil, resource.NewConfigValidationError(path, ErrInvalidFPort)
```
randhid, line 103: Please write an explanation on why these are error cases.

randhid (commented):
This is looking sweet!  A few more nits.

gateway/uplink.go
```diff
@@ -31,25 +31,34 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
+	if phyPayload[0] == confirmedUplinkMHdr {
+		uplinkType = "confirmed"
+	}
+	g.logger.Debugf("received %s uplink from %s", uplinkType, device.NodeName)
+
+	// confirmed data up, send ACK bit in downlink
+	sendAck := phyPayload[0] == confirmedUplinkMHdr
 
-	// we will send one device downlink from the do command per uplink.
+	var downlinkFramePayload []byte
```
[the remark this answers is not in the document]
oliviamiller: Either should be fine since in`createDownlink` we check that the payload is not nil before doing anything with it

gateway/uplink.go
```diff
@@ -31,25 +39,34 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 		return "", map[string]interface{}{}, errNoDevice
 	}
 
-	g.logger.Debugf("received uplink from %s", device.NodeName)
+	var uplinkType UplinkType = Unconfirmed
+	if phyPayload[0] == confirmedUplinkMHdr {
+		uplinkType = Confirmed
+	}
+	g.logger.Debugf("received %s uplink from %s", uplinkType, device.NodeName)
```
JohnN193, line 46: is this the only place that UplinkType gets used?

oliviamiller: right now yes

gateway/uplink.go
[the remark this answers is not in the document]
randhid: Yep just pointing out what the difference is, agree that either should be fine.

randhid (approved):

PR #36, opened 2025-03-13
[RSDK-10090] add waveshare hat model

JohnN193:
https://viam.atlassian.net/browse/RSDK-10090


.canon.yaml
```diff
@@ -1,5 +1,5 @@
 lorawan:
-    image_arm64: ghcr.io/viamrobotics/antique2:arm64-cache
+    image_arm64: ghcr.io/viamrobotics/rdk-devenv:arm64-cache
```
JohnN193, line 2: switched this to get testing to work locally

README.md
```diff
@@ -29,12 +29,12 @@ Hardware Required:
 
 See [Hardware Tested Section](https://github.com/viam-modules/lorawan?tab=readme-ov-file#hardware-tested) for hardware we have used this module with successfully.
 
-## Configure the `viam:sensor:sx1302-gateway`
+## Setup the `viam:sensor:sx1302-gateway`
 
 Navigate to the **CONFIGURE** tab of your machine in the [Viam app](https://app.viam.com/) and click the **+** button.
 [Add sx1302-gateway to your machine](https://docs.viam.com/operate/get-started/supported-hardware/#configure-hardware-on-your-machine).
```
JohnN193, line 36: moved this so it doesn't show up in markdown, when the config is already configured

README.md
```diff
@@ -55,6 +58,58 @@ The following attributes are available for `viam:sensor:sx1302-gateway` sensors:
...
 | power_en_pin | int | no | - | GPIO pin number for the power enable pin |
 
+## Setup the `viam:sensor:sx1302-hat-generic`
+
+Navigate to the **CONFIGURE** tab of your machine in the [Viam app](https://app.viam.com/) and click the **+** button.
+[Add sx1302-hat-generic to your machine](https://docs.viam.com/operate/get-started/supported-hardware/#configure-hardware-on-your-machine).
+
+## Configure the `viam:sensor:sx1302-hat-generic`
+
+Example gateway configuration:
```
JohnN193, line 68: technically should have an info blurb before this section but I couldn't think of what to write

meta.json
```diff
@@ -8,9 +8,21 @@
       {
           "api": "rdk:component:sensor",
           "model": "viam:lorawan:sx1302-gateway",
-          "markdown_link": "README.md#attributes",
+          "markdown_link": "README.md#configure-the-sx1302-gateway",
           "short_description": "A sensor component for a sx1302 LoRaWAN gateway HAT."
         },
```
JohnN193, line 13: may want to remove this from the meta.json now so users don't add it.

gateway/gateway.go
```diff
@@ -161,6 +179,22 @@ func NewGateway(
...
+		return cfg, nil
+	}
+	// check if we are the waveshare hat
+	waveshareHatCfg, err := resource.NativeConfig[*ConfigSX1302WaveshareHAT](conf)
+	if err != nil {
+		return nil, fmt.Errorf("the config %v does not match a supported config type", conf)
+	}
+	return waveshareHatCfg.getGatewayConfig(), nil
+}
+
```
JohnN193, line 197: in hardware testing & tests, this function worked. I couldn't trigger a situation where the `resource.Config` was getting written to the wrong struct. 

Can switch this to using enums with a switch if people prefer that.

.canon.yaml
oliviamiller: I originally changed this to antique so the module would work on both bullseye and bookworm, can you confirm this image will work with both? If it doesn't we can have a separate canon profile just for running tests

README.md
randhid:
```suggestion
## Configure the `viam:sensor:sx1302-hat-generic`

Example gateway configuration - note - the gpio pins MUST be set to the gpio pin numbers on your board connected to the reset and power-enable pins of the sx1302 hat to avoid a 15-minute reset loop.
```

README.md
```diff
@@ -55,6 +58,58 @@ The following attributes are available for `viam:sensor:sx1302-gateway` sensors:
...
+```json
+{
+    "board": "rpi",
+    "spi_bus": 0,
+    "reset_pin": 17,
+    "power_en_pin": 27
```
randhid, line 74:
```suggestion
    "reset_pin": 20000,
    "power_en_pin": 200000,
```
Nonsensical to not trip up the pins wrong.

README.md
```diff
@@ -55,6 +58,58 @@ The following attributes are available for `viam:sensor:sx1302-gateway` sensors:
...
+```
+
+### Attributes
+
+The following attributes are available for `viam:sensor:sx1302-waveshare-hat` sensors:
+
+| Name | Type | Required | Default | Description |
+|------|------|----------|---------|-------------|
+| board | string | yes | - | Name of the board connected to the HAT. The board communicates with the gateway through SPI |
+| spi_bus | int | no | 0 | SPI bus number (0 or 1) |
```
randhid, line 111:
```suggestion
| spi_bus | int | no | 0 | SPI bus number (0 or 1 on a raspberry pi) |
```

README.md
```diff
@@ -55,6 +58,58 @@ The following attributes are available for `viam:sensor:sx1302-gateway` sensors:
...
+
+### Attributes
+
+The following attributes are available for `viam:sensor:sx1302-hat-generic` sensors:
+
+| Name | Type | Required | Default | Description |
+|------|------|----------|---------|-------------|
+| board | string | yes | - | Name of the board connected to the HAT. The board communicates with the gateway through SPI |
+| reset_pin | int | yes | - | GPIO pin number for sx1302 reset pin |
+| spi_bus | int | no | 0 | SPI bus number (0 or 1) |
```
randhid, line 86:
```suggestion
| spi_bus | int | no | 0 | SPI bus number  |
```
jetson boards have 0 or 7 as spi buses I believe?

meta.json
```diff
@@ -8,9 +8,21 @@
...
+        {
+          "api": "rdk:component:sensor",
+          "model": "viam:lorawan:sx1302-hat-generic",
+          "markdown_link": "README.md#configure-the-sx1302-hat-generic",
+          "short_description": "A sensor component for a sx1302 LoRaWAN gateway HAT."
```
randhid, line 18:
```suggestion
          "short_description": "A sensor component for a sx1302 LoRaWAN gateway HAT connected by SPI to a single board computer."
```

gateway/gateway.go
randhid: We can just have a common provate creator get passed in the config pins as board objects, and an spi bus name? those are the three potential differences in code?

JohnN193: I do not understand what you mean. 

I needed this function because I wanted the waveshare_hat model to have a different json Config, so the current config validation functions wouldn't have to change.

gps does not require this function because they are an [always rebuild](https://github.com/viam-modules/gps/blob/7ca4edc68afd04213b5637ba3829a78c60201dec/rtk/gpsrtk.go#L40) model, but since we have `Reconfigure` we need to handle parsing different configs.

.canon.yaml
JohnN193: uhh do you have a bullseye I could test with? might just punt on this now that I know the trick lol

gateway/gateway.go
randhid: Ah yes Reconfigure means my plan doesn't work. I was thinking of having a common private function for both gateways that take in a reset pin and power enable pin as `board.GPIOPin` objects and a spi bus string, since that's the common functionality between the two.

But I did not consider reconfiguring, which takes in the two configuration objects. 

I like your enum idea a lot more in this case - it will be neat to implement a case structure on the Config type.

Thanks for the successful mind reading.

JohnN193: Kk updated. Realized I could use the model/model name in the resource.Config for this so I like where it landed

.canon.yaml
JohnN193: Kk so just using `rdk-devenv:arm64-cache` failed on bullseye.  added a second canon profile for testing

meta.json
randhid: Yeah let's do it!

oliviamiller (approved):

randhid (approved):

PR #38, opened 2025-03-18
RSDK-9323 Support Device time request downlink

oliviamiller:
The device sends an uplink in fopts requesting the current time, the module sends the deviceTimeans packet in a downlink. tested with em310 tilt, before we supported this it requested the time every few minutes, now that we are sending it it requests it again every 1-2 days.


gateway/downlink.go
```diff
@@ -131,34 +133,45 @@ func (g *gateway) createDownlink(device *node.Node, framePayload []byte, sendAck
...
+	if len(uplinkFopts) != 0 {
+		for _, b := range uplinkFopts {
+			switch b {
+			case deviceTimeCID:
+				g.logger.Debugf("got device time request from %s", device.NodeName)
+				deviceTimeAns, err := createDeviceTimeAns()
+				if err != nil {
+					g.logger.Errorf("failed to create device time answer: %w", err)
+				}
+				fopts = append(fopts, deviceTimeAns...)
```
JohnN193, line 148: so currently we only add to fopts if this specific command is called, but what other instances would we add/set the field? other requests from the uplink? can a user specify things to go into fopts?

gateway/downlink.go
```diff
@@ -211,3 +224,29 @@ func findDownLinkFreq(uplinkFreq int) int {
...
+	// Create frame payload
+	// Time is represented as seconds since GPS epoch
+	gpsEpoch := time.Date(1980, 1, 6, 0, 0, 0, 0, time.UTC)
+	now := time.Now()
+	secondsSinceGPSEpoch := uint32(now.Sub(gpsEpoch).Seconds())
+
+	buf := new(bytes.Buffer)
+	if err := binary.Write(buf, binary.LittleEndian, secondsSinceGPSEpoch); err != nil {
+		return nil, err
+	}
```
JohnN193, line 244: do we have any idea of when this would fail?

JohnN193 (approved):
LGTM, just some questions

gateway/downlink.go
oliviamiller: I would expect never but the linter fails if i dont check for the error

gateway/downlink.go
oliviamiller: there are other commands we could support, also some where we would send a request for something and the device would respond. In the future if we allow users to configure data rates/channels/etc then that stuff may go into the fopts.

oliviamiller: Table 5 in the lorawan spec doc has a list of all the possible commands

randhid (approved):

PR #39, opened 2025-03-19
RSDK-9331 Support linkcheckreq

oliviamiller:
Support the linkcheckreq by sending the linkcheckans back. The linkcheckans contains a margin calculation based on how far the SNR (signal to noise ratio) is from the minimum for demodulation, and the number of gateways that received the packet. The devices use this info to adject their data rate and RX power if needed to ensure the packet is able to be received by the gateway.

The min SNR for demodulation is based on the spread factor, added a check to not parse packets that are below the min SNR, as the data payload is these packets could be incorrect due to high noise.


gateway/uplink.go
```diff
@@ -54,9 +56,23 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
+	// Validate the MIC
+	mic, err := crypto.ComputeLegacyUplinkMIC(
+		types.AES128Key(device.NwkSKey), types.DevAddr(devAddrBE), uint32(frameCnt), phyPayload[:len(phyPayload)-4])
+	if err != nil {
+		return "", map[string]interface{}{}, err
+	}
+
+	if !bytes.Equal(phyPayload[len(phyPayload)-4:], mic[:]) {
+		return "", map[string]interface{}{}, errInvalidMIC
+	}
```
oliviamiller, line 74: didnt realize we werent doing this in here so added it, we should validate the MIC before parsing the data to ensure the data is valid, it's possible that adding this will help the dragino 300 degree readings problem but don't have a way of testing thats true.

JohnN193: is there anything else you want to add before the next release? if not lets make a new one after this pr, and we can bump pret to use the new changes to see if we get less "bad" data

oliviamiller: agree we should make a new release after this is in

gateway/gateway.go
```diff
@@ -405,19 +419,26 @@ func (g *gateway) receivePackets(ctx context.Context) {
 				if isDuplicate {
 					continue
 				}
+
+				minSNR := sfToSNRMin[int(packets[i].datarate)]
+				if float64(packets[i].snr) < minSNR {
+					g.logger.Debugf("packet skipped due to low signal noise ratio")
```
JohnN193, line 425: talked offline about maybe doing this check in other spots, or moving it to a `Warn`. olivia gonna try some things

gateway/uplink_test.go
```diff
@@ -5,74 +5,76 @@ import (
...
-		0x55, // FPORT
-		// Frame payload
-		0x00, 0x02, 0x25, 0x00,
-		0x2b, 0xc4, 0xdf, 0x00,
-		0x9c, 0x00, 0xaa, 0x00,
-		0x00, 0xbe,
-		// MIC
-		0x7d, 0xfe, 0x35, 0xfd,
+	// Create the plaintext payload that will decode to the expected values
+	plainText := []byte{
```
JohnN193, line 67: really hope these aren't too annoying to setup

JohnN193 (approved):
LGTM, only real question is the log level that we talked offline about

.canon.yaml
```diff
@@ -1,5 +1,4 @@
 lorawan:
-    default: true
```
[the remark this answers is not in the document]
JohnN193: o whoops

gateway/downlink.go
```diff
@@ -123,7 +125,8 @@ func accurateSleep(ctx context.Context, duration time.Duration) bool {
 // Downlink payload structure
 // | MHDR | DEV ADDR | FCTRL | FCNTDOWN |  FOPTS (optional)  |  FPORT | encrypted frame payload  |  MIC |
 // | 1 B  |   4 B    |  1 B  |    2 B   |       variable     |   1 B  |      variable            | 4 B  |
-func (g *gateway) createDownlink(device *node.Node, framePayload []byte, sendAck bool, uplinkFopts []byte) ([]byte, error) {
+func (g *gateway) createDownlink(device *node.Node, framePayload []byte, sendAck bool, uplinkFopts []byte, snr float64, sf int) (
+	[]byte, error) {
```
randhid, line 129: Remember you can group variables of similar types together like so for readability. Okay to leave as-is for now.
```suggestion
func (g *gateway) createDownlink(device *node.Node, framePayload, uplinkFopts []byte, sendAck bool, snr float64, sf int) (
	[]byte, error) {
```

randhid (approved):

PR #41, opened 2025-03-20
[RSDK-10290] check if nwk key and fcnt are set

JohnN193:
https://viam.atlassian.net/browse/RSDK-10290

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -19,6 +19,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -94,11 +95,11 @@ type Config struct {
 // deviceInfo is a struct containing OTAA device information.
 // This info is saved across module restarts for each device.
 type deviceInfo struct {
-	DevEUI   string `json:"dev_eui"`
-	DevAddr  string `json:"dev_addr"`
-	AppSKey  string `json:"app_skey"`
-	NwkSKey  string `json:"nwk_skey"`
-	FCntDown uint32 `json:"fcnt_down"`
+	DevEUI   string  `json:"dev_eui"`
+	DevAddr  string  `json:"dev_addr"`
+	AppSKey  string  `json:"app_skey"`
+	NwkSKey  string  `json:"nwk_skey"`
+	FCntDown *uint32 `json:"fcnt_down"`
 }
 
 func init() {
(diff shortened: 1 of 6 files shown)


node/node_utils.go
```diff
@@ -314,7 +314,7 @@ const (
 	// Minutes is minutes.
 	Minutes
 	// IntervalKey is the key for an interval DoCommand.
-	IntervalKey = "send_interval"
+	IntervalKey = "set_interval"
```
JohnN193, line 317: whoops. felt it was better to change this than to update the readmes, but honestly fine with the key being either of these

gateway/gateway.go
```diff
@@ -616,7 +617,12 @@ func (g *gateway) updateDeviceInfo(device *node.Node, d *deviceInfo) error {
 	device.AppSKey = appsKey
 	device.Addr = savedAddr
 	device.NwkSKey = nwksKey
-	device.FCntDown = d.FCntDown
+
+	// if we don't have an FCntDown in the device file, set it to a max number so we can tell.
+	device.FCntDown = math.MaxUint32
+	if d.FCntDown != nil {
+		device.FCntDown = *d.FCntDown
+	}
```
JohnN193, line 625: we prob don't have to checkboth the NwkSKey and FCntDown since the two of them were done at the same time. doesn't hurt to keep this tho

gateway/uplink.go
```diff
@@ -66,14 +67,21 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
 	if downlinkPayload != nil || foptsLength > 0 || sendAck {
-		payload, err := g.createDownlink(device, downlinkPayload, sendAck, fopts)
-		if err != nil {
-			return "", map[string]interface{}{}, fmt.Errorf("failed to create downlink: %w", err)
+		// devices that were configured on older versions of the module will not have these fields and cannot send downlinks.
+		// skip sending downlinks for those devices.
+		if len(device.NwkSKey) == 0 || device.FCntDown == math.MaxUint32 {
+			g.logger.Warnf("Sensor %v must be reset to support new features. "+
+				"Please physically restart the sensor to enable downlinks", device.NodeName)
+		} else {
```
JohnN193, line 75: I like being kinda vague with whats wrong here. LMK if I should change the log though

randhid: No this is great.

gateway/uplink.go
```diff
@@ -66,14 +67,21 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 	}
 
 	if downlinkPayload != nil || foptsLength > 0 || sendAck {
-		payload, err := g.createDownlink(device, downlinkPayload, sendAck, fopts)
-		if err != nil {
-			return "", map[string]interface{}{}, fmt.Errorf("failed to create downlink: %w", err)
+		// devices that were configured on older versions of the module will not have these fields and cannot send downlinks.
+		// skip sending downlinks for those devices.
+		if len(device.NwkSKey) == 0 || device.FCntDown == math.MaxUint32 {
+			g.logger.Warnf("Sensor %v must be reset to support new features. "+
```
randhid, line 73: Are we eventually going to sunset this? If so we can add a version number where we say we're no longer going to support the older version.

oliviamiller (approved):
lgtm!

gateway/uplink.go
JohnN193: I think once we know the current customers(modern, pret, maybeee eliot) will no longer hit this we should remove it. don't think its worth setting a hard date/version number tho.

randhid (approved):

PR #42, opened 2025-03-21
RSDK-10291 support EU868 band

oliviamiller:
Add support for EU868 band by adding a region_code attribute to gateway. Tested that it still works on the US band.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,9 @@ This module combines the functionality of a LoRaWAN gateway and network server, 
 It handles packet forwarding, device management, and message routing to allow LoRaWAN sensor data to be viewed and managed directly in Viam.
 This module provides the following models:
 
-- `sx1302-gateway`: Sensor model for a SX1302 LoRaWAN concentrator hat connected to a Raspberry Pi.
-- `node`: Sensor model for any class A, US915 LoraWAN end device.
+- `sx1302-waveshare-hat`: Sensor model for the [waveshare loRaWAN gateway hat](https://www.waveshare.com/wiki/SX1302_LoRaWAN_Gateway_HAT)
+- `sx1302-hat-generic`: Sensor model for other SX1302 LoRaWAN concentrator hats.
+- `node`: Sensor model for any class A, US915/EU868 LoraWAN end device.
 - `dragino-LHT65N`: Sensor model for the dragino LHT65N temperature and humidity sensor.
 - `milesight-ct101`: Sensor model for the milesight ct101 current transformer.
 - `milesight-em310-tilt`: Sensor model for the milesight em310 tilt sensor.
@@ -16,7 +17,7 @@ This module provides the following models:
 You'll configure the `sx1302-gateway` model, and one or more `node`s depending on how many sensors you have.
 
 Compatible with:
-- US915 frequency band
+- US915 or EU868 frequency band
 - Class A Devices
 - LoraWAN MAC version 1.0.3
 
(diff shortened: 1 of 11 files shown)


gateway/gateway.go
```diff
@@ -152,6 +216,11 @@ func (conf *Config) Validate(path string) ([]string, error) {
 	}
 	deps = append(deps, conf.BoardName)
 
+	if conf.Region != "" && !strings.Contains(conf.Region, "US") && !strings.Contains(conf.Region, "EU") &&
```
oliviamiller, line 219: want the user to be able to just put "eu" or "868"  and have it still work, theres prob a better way to do this if anyone has suggestions

JohnN193 (commented):
First pass, mostly just moving some code around.

gateway/regions.go
```diff
@@ -0,0 +1,73 @@
...
+	region = strings.ToUpper(region)
+	switch region {
+	case "US", "US915", "915":
+		return true
+	case "EU", "EU868", "868":
+		return true
+	default:
+		return false
+	}
+}
```
JohnN193, line 73: instead of returning a bool, return the region. have default return unspecified

gateway/gateway.go
```diff
@@ -288,7 +305,20 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		return fmt.Errorf("error initializing the gateway: %w", err)
 	}
 
-	errCode := C.setUpGateway(C.int(cfg.Bus))
+	region := US
+	if strings.Contains(cfg.Region, "EU") || strings.Contains(cfg.Region, "868") {
+		g.logger.Infof("configuring for EU868 region")
+		region = EU
+	}
```
JohnN193, line 312: reuse the isValidRegion that returns the region code. then if unspecified, set to default/US

gateway/gateway.go
```diff
@@ -152,6 +162,12 @@ func (conf *Config) Validate(path string) ([]string, error) {
 	}
 	deps = append(deps, conf.BoardName)
 
+	if conf.Region != "" {
+		if !isValidRegion(conf.Region) {
```
JohnN193, line 166: `isValidRegion == 0` if using the unspecified region

JohnN193 (approved):
so pretty nit, but added a bunch of thoughts for using the isValidRegion function abit more. Very optional and LGTM otherwise

README.md
```diff
@@ -57,6 +59,7 @@ The following attributes are available for `viam:sensor:sx1302-gateway` sensors:
 | reset_pin | int | yes | - | GPIO pin number for sx1302 reset pin |
 | spi_bus | int | no | 0 | SPI bus number (0 or 1) |
 | power_en_pin | int | no | - | GPIO pin number for the power enable pin |
+| region_code | string | no | US915 | frequency region of your gateway (US915 or EU868) |
```
randhid, line 62: Is there a default if it is not required?

gateway/gateway.go
```diff
@@ -288,7 +306,15 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
 
-	errCode := C.setUpGateway(C.int(cfg.Bus))
+	region := getRegion(cfg.Region)
+	switch region {
+	case US, unspecified:
+		g.regionInfo = regionInfoUS
+	case EU:
+		g.regionInfo = regionInfoEU
+
+	}
```
randhid, line 316: Do we have an info log anywhere telling us what region we built the lorawan gateway/sensors with?

For the future - node/gateway region mismatches of region are possible in my reading of this pr. Addressing this properly will likely be easier with one module-level lorawan library that can bookkeep information about the node and gateway.

randhid (approved):
One request - Is there an info log telling us what region we've built the node and sensor with?

README.md
oliviamiller: yes default is US915

gateway/gateway.go
oliviamiller: I will add an info log! The module doesn't know what what region the nodes are, its on the user to make sure they match their nodes and gateway.

PR #45, opened 2025-03-22
[RSDK-10308] do not validate mic if no key

JohnN193:
https://viam.atlassian.net/browse/RSDK-10308

### DIFF
--- a/gateway/uplink.go
+++ b/gateway/uplink.go
@@ -64,7 +64,7 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 	frameCnt := binary.LittleEndian.Uint16(phyPayload[6:8])
 
 	// only validate the MIC if we have a NwkSKey set
-	if len(device.NwkSKey) == 0 {
+	if len(device.NwkSKey) != 0 {
 		// Validate the MIC
 		mic, err := crypto.ComputeLegacyUplinkMIC(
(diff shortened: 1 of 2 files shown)


gateway/uplink.go
```diff
@@ -63,15 +63,18 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
 	frameCnt := binary.LittleEndian.Uint16(phyPayload[6:8])
 
-	// Validate the MIC
-	mic, err := crypto.ComputeLegacyUplinkMIC(
-		types.AES128Key(device.NwkSKey), types.DevAddr(devAddrBE), uint32(frameCnt), phyPayload[:len(phyPayload)-4])
-	if err != nil {
-		return "", map[string]interface{}{}, err
-	}
+	// only validate the MIC if we have a NwkSKey set
+	if len(device.NwkSKey) == 0 {
```
oliviamiller, line 67: shouldn't this be if != 0 ?

JohnN193: O true

JohnN193, line 67:
```suggestion
	if len(device.NwkSKey) != 0 {
```

oliviamiller (approved):
lgtm! thanks for the quick fix

randhid (approved):
Thanks for such a quick fix!

JohnN193:
> Thanks for such a quick fix!

np, but we still want eliot/others to restart their sensors at some point so they can utilize the new features

PR #48, opened 2025-03-25
Check for invalid payload lengths

oliviamiller:
Got a panic when a join request didn't contain the expected data, adding checks to error if a payload doesn't have the expected bytes.

### DIFF
--- a/gateway/join.go
+++ b/gateway/join.go
@@ -61,6 +61,9 @@ func (g *gateway) handleJoin(ctx context.Context, payload []byte, packetTime tim
 // | 1 B  |   8 B    |    8 B   |     2 B      |  4 B  |
 // https://lora-alliance.org/wp-content/uploads/2020/11/lorawan1.0.3.pdf page 34 for more info on join request.
 func (g *gateway) parseJoinRequestPacket(payload []byte) (joinRequest, *node.Node, error) {
+	if len(payload) != 23 {
+		return joinRequest{}, nil, errInvalidLength
+	}
 	var joinRequest joinRequest
 
 	// everything in the join request payload is little endian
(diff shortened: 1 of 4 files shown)


gateway/join.go
```diff
@@ -61,6 +61,9 @@ func (g *gateway) handleJoin(ctx context.Context, payload []byte, packetTime tim
 // | 1 B  |   8 B    |    8 B   |     2 B      |  4 B  |
 // https://lora-alliance.org/wp-content/uploads/2020/11/lorawan1.0.3.pdf page 34 for more info on join request.
 func (g *gateway) parseJoinRequestPacket(payload []byte) (joinRequest, *node.Node, error) {
+	if len(payload) != 23 {
+		return joinRequest{}, nil, errInvalidLength
+	}
```
randhid, line 66: Add a comment why we do this check first.

gateway/uplink.go
```diff
@@ -24,13 +25,19 @@ const (
...
 // | MHDR | DEV ADDR|  FCTL |   FCnt  | FPort   |  FOpts     |  FRM Payload | MIC |
 // | 1 B  |   4 B    | 1 B   |  2 B   |   1 B   | variable    |  variable   | 4B  |
 // Returns the node name, readings and error.
 func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packetTime time.Time, snr float64, sf int) (
 	string, map[string]interface{}, error,
 ) {
+	if len(phyPayload) < 13 {
+		return "", map[string]interface{}{}, errInvalidLength
+	}
+
```
randhid, line 40: Please `erros.Join()` the two returned errors in this file with a descriptive error that tells us in what part of the stack we're seeing the invalid payloads, for better insights when they do happen.

randhid (approved):

JohnN193 (approved):
lgtm, only thought is maybe getting rid of magic numbers, but since these are pretty specific & we have comments I think they are fine

PR #50, opened 2025-03-27
add sqlite

JohnN193:


gateway/downlink.go
```diff
@@ -177,11 +177,8 @@ func (g *gateway) createDownlink(device *node.Node, framePayload, uplinkFopts []
 		FCntDown: &device.FCntDown,
 	}
 
-	if err = g.searchAndRemove(g.dataFile, device.DevEui); err != nil {
-		return nil, fmt.Errorf("failed to remove device info from file: %w", err)
-	}
-	if err = g.addDeviceInfoToFile(g.dataFile, deviceInfo); err != nil {
-		return nil, fmt.Errorf("failed to add device info to file: %w", err)
+	if err = g.insertOrUpdateDeviceInDB(context.Background(), deviceInfo); err != nil {
```
oliviamiller, line 180: should we pass in a ctx into createDownlink to use here instead of` context.Background()` ?

gateway/gateway.go
```diff
@@ -173,8 +172,7 @@ type gateway struct {
 
 	logReader  *os.File
 	logWriter  *os.File
-	dataFile   *os.File
-	dataMu     sync.Mutex
+	db         *sql.DB
```
oliviamiller, line 175: add a comment here on what the DB is used for

gateway/gateway.go
```diff
@@ -479,6 +469,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 	if _, ok := cmd["validate"]; ok {
 		return map[string]interface{}{"validate": 1}, nil
 	}
+	if deviceReq, ok := cmd["return_devices"]; ok {
```
oliviamiller, line 472: would a user ever use this or just for internal testing?

gateway/gateway.go
```diff
@@ -479,6 +469,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+		resp := map[string]interface{}{}
+		deviceStr, _ := deviceReq.(string)
+		// look for a specific device if requested. If the device name is not a string we just return all devices.
+		if deviceStr != "" {
+			// check if we are receiving data from the device.
```
oliviamiller, line 477:
```suggestion
			// check if the device has joined the network.
```

gateway/gateway.go
```diff
@@ -479,6 +469,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+			return nil, fmt.Errorf("device %v is not currently joined", deviceStr)
+		}
+		resp["all devices"] = deviceStr
+		// Read the device info from the file
+		devices, err := g.getAllDevicesFromDB(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read device info from db: %w", err)
+		}
+		for _, device := range devices {
+			resp[device.DevEUI] = device
```
oliviamiller, line 491: we could also save the nodename to the db and return the name instead of devEUI here, would make it easier to find the device youre looking for. But might be more complicated if the user changes their sensor's name so dont have to do

gateway/gateway_test.go
```diff
@@ -473,68 +451,68 @@ func TestStartCLogging(t *testing.T) {
...
-	device := g.devices[testNodeName]
-
-	err := g.updateDeviceInfo(device, validInfo)
-	test.That(t, err, test.ShouldBeNil)
-	test.That(t, device, test.ShouldNotBeNil)
-	test.That(t, device.NodeName, test.ShouldEqual, testNodeName)
-	test.That(t, device.AppSKey, test.ShouldResemble, newAppSKey)
-	test.That(t, device.Addr, test.ShouldResemble, newDevAddr)
-}
+// func TestSearchForDeviceInFile(t *testing.T) {
```
oliviamiller, line 454: reminder to remove this before merge

gateway/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+	errNoDeviceInDB = errors.New("error device not found")
+	errNoDB         = errors.New("error device file not found")
+)
+
+// Create or open a sqlite db file used to save device data across restarts.
+func (g *gateway) setupSqlite(ctx context.Context) error {
+	moduleDataDir := os.Getenv("VIAM_MODULE_DATA")
+
+	var txtDevices []deviceInfo
+	// check if the machine has an old devicedata file for us to migrate
```
oliviamiller, line 30: I think if you should check if the .db file already exists before doing this in case the gateway is just being rebuilt

gateway/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	return nil
+}
+
+func (g *gateway) insertOrUpdateDeviceInDB(ctx context.Context, device deviceInfo) error {
+	if g.db == nil {
+		return errNoDB
+	}
+	_, err := g.db.ExecContext(ctx, "insert or replace into devices (devEui, appSKey, nwkSKey, devAddr, fCntDown) VALUES(?, ?, ?, ?, ?);",
```
oliviamiller, line 75: if its the same devEUI but everything else is different this will replace the entry right?

gateway/sqlite.go
```diff
@@ -0,0 +1,165 @@
...
+
+	//nolint:errcheck
+	defer rows.Close()
+
+	devices := []deviceInfo{}
+	for rows.Next() {
+		if ctx.Err() != nil {
+			return []deviceInfo{}, ctx.Err()
+		}
+		// var devEuiOut, appSKeyOut, nwkSKeyOut, devAddrOut string
```
oliviamiller, line 121: delete comments

gateway/downlink.go
JohnN193: added!

gateway/gateway.go
JohnN193: a user would use this to see all devices that have ever joined their gateway. mostly a debugging tool tho.

gateway/gateway.go
JohnN193: so I think I still want to search by devEUI for now, but I will go ahead and add node name to the table

gateway/sqlite.go
JohnN193: yep!

gateway/gateway.go
```diff
@@ -104,6 +103,7 @@ type deviceInfo struct {
 	AppSKey  string  `json:"app_skey"`
 	NwkSKey  string  `json:"nwk_skey"`
 	FCntDown *uint16 `json:"fcnt_down"`
+	NodeName string  `json:"node_name"`
```
randhid, line 106: A little confused why we're adding this as a singular string - couldn't there be multiple nodes per gateway?

JohnN193: We have multiple devices per gateway, and each device only has one name technically

gateway/gateway.go
```diff
@@ -191,19 +190,11 @@ func NewGateway(
 		started: false,
 	}
 
-	// Create or open the file used to save device data across restarts.
-	moduleDataDir := os.Getenv("VIAM_MODULE_DATA")
-	filePath := filepath.Join(moduleDataDir, "devicedata.txt")
-	//nolint:gosec
-	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE, 0o600)
-	if err != nil {
```
randhid, line 199: Where is the creation logic here? Does `os.OpenFile` create a file if it does not exist in the directory?

gateway/join_test.go
```diff
@@ -378,10 +278,13 @@ func TestHandleJoin(t *testing.T) {
...
 
 	g := &gateway{
-		logger:   logging.NewTestLogger(t),
-		devices:  devices,
-		dataFile: createDataFile(t),
+		logger:  logging.NewTestLogger(t),
+		devices: devices,
 	}
+	dataDirectory1 := t.TempDir()
+	t.Setenv("VIAM_MODULE_DATA", dataDirectory1)
```
nicksanford, line 285: Rather than doing this `t.Setenv` here, you could have the `g.setupSqlite` take a second parameter of `pathPrefix string` and not need to stub out the environment variable.

gateway/sqlite.go
```diff
@@ -0,0 +1,167 @@
...
+	devices(devEui TEXT NOT NULL PRIMARY KEY, appSKey TEXT, nwkSKey TEXT, devAddr TEXT, fCntDown INTEGER, nodeName TEXT);
+	`
+	sqlStmt := cmd + keys
+	if _, err = db.ExecContext(ctx, sqlStmt); err != nil {
+		return err
+	}
+	g.db = db
+
+	// check if the machine has an old devicedata file for us to migrate
+	filePathTXT := filepath.Join(moduleDataDir, "devicedata.txt")
```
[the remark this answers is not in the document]
nicksanford: Also, can this migration scrpt fail at any point and succeed cleanly on subsequent retries?

gateway/gateway.go
```diff
@@ -479,6 +470,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
 	}
+	if deviceReq, ok := cmd["return_devices"]; ok {
+		resp := map[string]interface{}{}
+		deviceStr, _ := deviceReq.(string)
+		// look for a specific device if requested. If the device name is not a string we just return all devices.
+		if deviceStr != "" {
+			// check if the device has joined the network.
+			resp[deviceStr], ok = g.devices[deviceStr]
+			if ok {
+				resp["specific device currently joined"] = deviceStr
```
nicksanford, line 481: why do we set this only to inevitably not use this on line 483?

gateway/gateway.go
```diff
@@ -479,6 +470,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 	if _, ok := cmd["validate"]; ok {
 		return map[string]interface{}{"validate": 1}, nil
 	}
+	if deviceReq, ok := cmd["return_devices"]; ok {
+		resp := map[string]interface{}{}
+		deviceStr, _ := deviceReq.(string)
+		// look for a specific device if requested. If the device name is not a string we just return all devices.
+		if deviceStr != "" {
+			// check if the device has joined the network.
+			resp[deviceStr], ok = g.devices[deviceStr]
```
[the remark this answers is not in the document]
nicksanford: is g.devices modified / read by another goroutines? If so, what mutex is protecting it?

gateway/gateway.go
```diff
@@ -479,6 +470,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+			if ok {
+				resp["specific device currently joined"] = deviceStr
+			}
+			return nil, fmt.Errorf("device %v is not currently joined", deviceStr)
+		}
+		resp["all devices"] = deviceStr
```
nicksanford, line 485: this is going to equal `{"all devices": ""}` right? What does that mean?

gateway/sqlite.go
```diff
@@ -0,0 +1,167 @@
...
+		devEui).Scan(&newDevice.DevEUI, &newDevice.AppSKey, &newDevice.NwkSKey,
+		&newDevice.DevAddr, &newDevice.FCntDown, &newDevice.NodeName); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return &deviceInfo{}, errNoDeviceInDB
+		}
+		return &deviceInfo{}, err
+	}
+	return &newDevice, nil
+}
+
```
nicksanford, line 103: I recommend not returning pointers if you use plane old structs elsewhere.
```suggestion
func (g *gateway) findDeviceInDB(ctx context.Context, devEui string) (deviceInfo, error) {
	if g.db == nil {
		return nil, errNoDB
	}
	var zero deviceInfo
	devEui = strings.ToUpper(devEui)
	newDevice := deviceInfo{}
	if err := g.db.QueryRowContext(ctx, "select * from devices where devEui = ?",
		devEui).Scan(&newDevice.DevEUI, &newDevice.AppSKey, &newDevice.NwkSKey,
		&newDevice.DevAddr, &newDevice.FCntDown, &newDevice.NodeName); err != nil {
		if errors.Is(err, sql.ErrNoRows) {
			return zero, errNoDeviceInDB
		}
		return zero, err
	}
	return newDevice, nil
}

```

gateway/sqlite.go
```diff
@@ -0,0 +1,167 @@
...
+	return err
+}
+
+func (g *gateway) findDeviceInDB(ctx context.Context, devEui string) (*deviceInfo, error) {
+	if g.db == nil {
+		return nil, errNoDB
+	}
+	devEui = strings.ToUpper(devEui)
+	newDevice := deviceInfo{}
+	if err := g.db.QueryRowContext(ctx, "select * from devices where devEui = ?",
```
nicksanford, line 93: do you not need a semicolon at the end of the query?

gateway/sqlite.go
```diff
@@ -0,0 +1,167 @@
...
+		return &deviceInfo{}, err
+	}
+	return &newDevice, nil
+}
+
+func (g *gateway) getAllDevicesFromDB(ctx context.Context) ([]deviceInfo, error) {
+	if g.db == nil {
+		return nil, errNoDB
+	}
+	queryAll := `SELECT * FROM devices`
```
nicksanford, line 108: do you not need a semicolon at the end of the query?

gateway/sqlite.go
```diff
@@ -0,0 +1,167 @@
...
+			return nil, err
+		}
+
+		devices = append(devices, device)
+	}
+	return devices, nil
+}
+
+// Function to read the device info from the persitent data file.
+func readFromFile(filePath string) ([]deviceInfo, error) {
```
nicksanford, line 137:
```suggestion
func readFromJsonFile(filePath string) ([]deviceInfo, error) {
```

gateway/sqlite.go
```diff
@@ -0,0 +1,167 @@
...
+	filePath = filepath.Clean(filePath)
+	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE, 0o600)
+	if err != nil {
+		return nil, err
+	}
+	//nolint:errcheck
+	defer file.Close()
```
nicksanford, line 144: log error

gateway/sqlite_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"testing"
+
+	"go.viam.com/rdk/logging"
+	"go.viam.com/test"
+)
+
+func TestSetupSqlite(t *testing.T) {
+	t.Run("Good setup with no previous db", func(t *testing.T) {
+		dataDirectory1 := t.TempDir()
+		t.Setenv("VIAM_MODULE_DATA", dataDirectory1)
```
nicksanford, line 19: I recommend changing the code to not require stubbing out this env var.

gateway/gateway_test.go
```diff
@@ -579,9 +527,8 @@ func TestClose(t *testing.T) {
 		_, err = g.logReader.Read(buf)
 		test.That(t, err, test.ShouldNotBeNil)
 	}
-	if g.dataFile != nil {
-		buf := make([]byte, 1)
-		_, err = g.dataFile.Read(buf)
+	if g.db != nil {
```
nicksanford, line 530: isn't this impossible?

JohnN193: yeah, tbh was just using the same setup that was here, but none of these should be `nil` in this test

gateway/gateway.go
JohnN193: yeah, when given the os.O_CREATE flag os.OpenFile will create a file if it does not exist. 

tbh I should remove the O_CREATE flag, since we now only use this function if we already know the txt file exists

gateway/gateway.go
```diff
@@ -479,6 +470,29 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
 	if _, ok := cmd["validate"]; ok {
 		return map[string]interface{}{"validate": 1}, nil
 	}
+	if deviceReq, ok := cmd["return_devices"]; ok {
+		resp := map[string]interface{}{}
+		deviceStr, _ := deviceReq.(string)
+		// look for a specific device if requested. If the device name is not a string we just return all devices.
```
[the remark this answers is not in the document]
JohnN193: so by swallowing the ok check I know that deviceStr should either be a valid string or `""`. 

I chose not to return an error if its not a string because I wanted the docommand to still return all devices if it was used incorrectly.

gateway/gateway.go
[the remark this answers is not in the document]
JohnN193: tbh missed that this part of the DoCommand wasn't actually working. gonna rework or remove.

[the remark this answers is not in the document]
JohnN193: `g.devices` gets read from in the receive packets thread, with individual devices also being read/writen to via pointers. 

no mutex protects it in the other goroutine

gateway/gateway.go
JohnN193: I missed that this part of the DoCommand was no longer functional. I ended up only using/testing the all devices path.

gateway/gateway.go
JohnN193: was intended to signal to the user that we were in the all devices path. going to remove the individual sensor logic

gateway/sqlite.go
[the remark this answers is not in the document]
JohnN193: not that I am aware of. My understanding is this should(probably) only fail if the json file is corrupted in some way, in which case the state cannot be recovered without someone manually fixing the problem. 

manually fixing the problem involves either
a. fix the json file
b. remove the json file and reset all of the sensors

gateway/sqlite.go
JohnN193: so nothing yelled at me for needing a semicolon, but if its technically correct formatting for sqlite I'd be happy to add them in

gateway/sqlite_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"go.viam.com/rdk/logging"
+	"go.viam.com/test"
+)
+
```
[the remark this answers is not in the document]
JohnN193: when i tried this it failed because of cgo

gateway/gateway.go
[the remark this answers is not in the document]
JohnN193: made a ticket to do more mutexing https://viam.atlassian.net/browse/RSDK-10362

gateway/gateway.go
JohnN193: removed the deviceStr logic

gateway/gateway.go
```diff
@@ -714,9 +704,9 @@ func (g *gateway) Close(ctx context.Context) error {
 		delete(loggingRoutineStarted, g.Name().Name)
 	}
 
-	if g.dataFile != nil {
-		if err := g.dataFile.Close(); err != nil {
-			g.logger.Errorf("error closing data file: %s", err)
+	if g.db != nil {
+		if err := g.db.Close(); err != nil {
```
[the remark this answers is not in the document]
JohnN193: 1. calling the Close function twice did not appear to cause any issues
2. I don't believe this is possible with viam's stack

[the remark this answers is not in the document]
randhid: 2. Is not currently possible/ if you can call ANY API on a `Close`d resource sound the alarm to cheuk. That is a resource graph bug, not a feature.

gateway/sqlite.go
```diff
@@ -0,0 +1,176 @@
...
+		devices = append(devices, device)
+	}
+
+	return devices, nil
+}
+
+// Function to migrate the device info from the persitent data file into a sqlite db
+func (g *gateway) migrateDevicesFromJSONFile(ctx context.Context, pathPrefix string) error {
+	// check if the machine has an old devicedata file for us to migrate
+	filePathTXT := filepath.Join(pathPrefix, "devicedata.txt")
```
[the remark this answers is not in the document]
JohnN193: yeah, its the same folder we were using before. pathPrefix points to the `VIAM_MODULE_DATA` env var. Ideally we can remove this in abit, but I do see a world where we keep this in as a quicker way to migrate some sensors(i.e. the safelease test case)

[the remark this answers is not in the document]
randhid: Edited your comment since this repo is visible to everyone. Okay!

randhid (approved):

gateway/sqlite.go
```diff
@@ -0,0 +1,177 @@
...
+	if g.db == nil {
+		return nil, errNoDB
+	}
+	queryAll := `SELECT * FROM devices;`
+	rows, err := g.db.QueryContext(ctx, queryAll)
+	if err != nil {
+		if errors.Is(err, errDBClosedInternal) {
+			return nil, errDBClosed
+		}
+		if err.Error() == errDBClosedInternal.Error() {
```
oliviamiller, line 99: why do you need both of these

oliviamiller (approved):
lgtm, just one question

gateway/sqlite.go
JohnN193: i don't, I thought I deleted all instances of the `errors.Is` case whoops

gateway/sqlite.go
```diff
@@ -0,0 +1,182 @@
...
+	var zero deviceInfo
+	devEui = strings.ToUpper(devEui)
+	newDevice := deviceInfo{}
+	if err := g.db.QueryRowContext(ctx, "select * from devices where devEui = ?;",
+		devEui).Scan(&newDevice.DevEUI, &newDevice.AppSKey, &newDevice.NwkSKey,
+		&newDevice.DevAddr, &newDevice.FCntDown, &newDevice.NodeName, &newDevice.MinUplinkInterval); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return zero, errNoDeviceInDB
+		}
+		if err.Error() == errDBClosedInternal.Error() {
```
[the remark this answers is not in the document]
JohnN193: this one I need, in the other case errors.Is was checking against `errDBClosedInternal`, but I couldn't do that since the error sql returns something different technically, even though the message is the same. 

here we check an error that actually returns from sql, in addition to the `errDBClosedInternal` error

PR #51, opened 2025-03-31
RSDK-10335 enforce duty cycle regulation on EU devices

oliviamiller:
Sends a DutycycleReq to EU devices to limit the duty cycle. The device will limit its uplinks based on the num in this request. The device can send a dutycycleAns but according to docs this is not required to be sent by the devices. 

Logs a warning with an estimation of the minimum uplink interval based on the uplink's size and spread factor. For the dragino sensors sending at a SF of 10 and payload size of 24 bytes, the min interval is 47 seconds. Also saves this number to the node struct and logs a warning if the user tries to set an interval below the min.

Confirmed US gateway still works, wasn't able to test the dutyCycleReq in the US


gateway/downlink.go
```diff
@@ -241,3 +249,12 @@ func createLinkCheckAns(snr float64, sf int) []byte {
 
 	return payload
 }
+
+// EU devices have limitations on how often they can use the frequency band.
+// duty cycle = 1/2^(MaxDCycle).
+func createDutyCycleReq() []byte {
+	payload := make([]byte, 0)
+	payload = append(payload, dutyCycleCID)
+	payload = append(payload, 0x07) // 2^-7 = 0.78% duty cyle, closest to 1% we can get.
```
oliviamiller, line 258: This request is very limited with the duty cycles it can request, the closest we can get to set a 1% duty cycle is 0.78%.

gateway/downlink.go
```diff
@@ -10,22 +10,20 @@ import (
...
 	"go.thethings.network/lorawan-stack/v3/pkg/crypto"
 	"go.thethings.network/lorawan-stack/v3/pkg/types"
 	"go.viam.com/utils"
 )
 
 const (
-	joinDelay      = 6         // rx2 delay in seconds for sending join accept message.
-	downlinkDelay  = 2         // rx2 delay in seconds for downlink messages.
-	rx2FrequencyUS = 923300000 // rx2 default freq for US
-	rx2FrequencyEU = 869525000 // rx2 default freq for EU region
```
oliviamiller, line 22: missed removing these in my last PR, they were moved to regions package

node/node.go
```diff
@@ -265,9 +269,20 @@ func getGateway(ctx context.Context, deps resource.Dependencies) (sensor.Sensor,
 	if !ok {
 		return nil, errors.New("dependency must be the sx1302-gateway sensor")
 	}
-	if retVal.(float64) != 1 {
-		return nil, errors.New("dependency must be the sx1302-gateway sensor")
+	re, ok := retVal.(float64)
```
JohnN193, line 272: could this be retVal.(regions.Region)? no idea if that works

node/node_utils.go
```diff
@@ -345,6 +346,18 @@ func (n *Node) SendIntervalDownlink(ctx context.Context, req IntervalRequest) (m
...
+		n.logger.Warnf(`the duty cycle limit in the EU region is 1%%. Ensure your
+		 uplink interval complies with this restriction to avoid transmission issues`)
+	}
+
+	if n.MinIntervalSeconds != 0 {
+		if n.MinIntervalSeconds > req.IntervalMin*60 {
+			n.logger.Warnf(`requested uplink interval (%.2f minutes) may exceed the legal duty cycle limit.
+			The device may not transmit at this interval`, req.IntervalMin)
+		}
+	}
```
JohnN193, line 359: where does this field get set?

gateway/downlink.go
JohnN193: thanks i hate it

node/node_utils.go
oliviamiller: `parseDataUplink` line 108

oliviamiller: not sure I understand what you mean, the docommand here isn't changing any fields, but the gateway is setting this field for the device in parseDataUplink.

oliviamiller: ah wait yeah this wouldn't work

JohnN193: parseDataUplink happens in the receive thread of `gateway.go`, in gateway modules. 

SendIntervalDownlink is called by nodes, via their `DoCommand` or node `reconfigure`

so the only way for parseDataUplink to actually set the field is if the pointer/struct we send during `register_device` points back to the node

if it does work does it also mean

- the node should have access to the apps/nwks keys
- the another gateway would also have access to those keys, via the same pointers
- our setup has more threads that can access information than we thought

node/node.go
JohnN193: kk was worth a shot

node/node_utils.go
JohnN193: o i refreshed & u commented

also for testing lets go ahead and enforce the EU requirement on our NA setups. we can remove it before merging

oliviamiller: I tried to do that but sending the dutyCycleReq to the US device caused it to start resending join requests, im guessing it doesn't expect to get that command and freaks out

JohnN193: ah kk. then lets ping felix to see if we can test with his setup

gateway/uplink.go
```diff
@@ -294,3 +304,39 @@ func (g *gateway) getFOptsToSend(fopts []byte, device *node.Node) []byte {
...
+	codeRate := 1               // lora packets using 4/5 coderate so use 1
+	crcNumBits := 16            // num bits in crc
+
+	payloadBits := 8 * size // get the number of bits of the payload
+	payloadBits += crcNumBits
+	payloadBits -= 4 * (sf) // Unclear what the -4 does, not explained in datasheet.
+	payloadBits += 8        // Mystery number from datasheet formula
+	payloadBits += explicitHeaderSymbols
+
+	bitsPerSymbol := float64(sf)
```
randhid, line 327: Can this ever be 0?

gateway/gateway.go
```diff
@@ -475,9 +480,9 @@ func (g *gateway) updateReadings(name string, newReadings map[string]interface{}
 func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (map[string]interface{}, error) {
 	g.mu.Lock()
 	defer g.mu.Unlock()
-	// Validate that the dependency is correct.
+	// Validate that the dependency is correct, returns the gateway's region.
 	if _, ok := cmd["validate"]; ok {
-		return map[string]interface{}{"validate": 1}, nil
+		return map[string]interface{}{"validate": g.region}, nil
```
randhid, line 485: This is kinda nice.

gateway/downlink.go
randhid: This sucks.

gateway/downlink_test.go
```diff
@@ -187,22 +209,27 @@ func TestCreateDownlink(t *testing.T) {
 				test.That(t, payload[5], test.ShouldEqual, tt.expectedFctrl)
 
 				currentPos := 8 // Start after MHDR(1) + DevAddr(4) + FCtrl(1) + FCnt(2)
-				if tt.uplinkFopts != nil {
+				if tt.expectedFOptsLength != 0 {
+					fOpts := payload[8 : 8+tt.expectedFOptsLength]
 					for _, b := range tt.uplinkFopts {
 						if b == deviceTimeCID {
-							test.That(t, payload[currentPos], test.ShouldEqual, deviceTimeCID)
+							test.That(t, fOpts, test.ShouldContain, uint8(deviceTimeCID))
```
randhid, line 216: Is there a case where the `fOpts` contains the uint8 we're searching for but in the wrong position in the fOpts array?

node/node_utils.go
randhid: I'd suggest making an rc with just this code and release it to him to test out any issues. Does that sound like a good plan?

gateway/uplink.go
oliviamiller: no, the spread factor sent by the device will be 4-12

gateway/downlink_test.go
oliviamiller: yea that would be possible, updated the test to check for the right position

oliviamiller:
Updated this PR to add the pollGateway function which will call a docommand on the gateway to return the device info from the DB, and then updates those fields in the node struct. This is how we are getting an updated MinInterval and can use it for sharing keys in the future.

draginolht65n/sensor.go
```diff
@@ -107,6 +108,8 @@ func newLHT65N(
 		return nil, err
 	}
 
+	n.node.Workers = utils.NewBackgroundStoppableWorkers(n.node.PollGateway)
```
JohnN193, line 111: can this live in the generic node code? potentially in reconfigure since the gateways we talk to can change

node/node.go
```diff
@@ -235,7 +242,7 @@ func (n *Node) Reconfigure(ctx context.Context, deps resource.Dependencies, conf
 }
 
 // getGateway sends the validate docommand to the gateway to confirm the dependency.
-func getGateway(ctx context.Context, deps resource.Dependencies) (sensor.Sensor, error) {
+func (n *Node) getGateway(ctx context.Context, deps resource.Dependencies) (sensor.Sensor, error) {
```
JohnN193, line 245: either we should return both the region and the gateway, or store them both on the struct within this function

draginolht65n/sensor.go
JohnN193: we can have this live in reconfigure, then if we detect the gateway has changed we stop the worker and restart the thread. 

main worry is that they both use the same mutex, so I don't know how we are supposed to cancel the background worker while we hold the mutex in reconfigure

JohnN193 (commented):
Main concern is setting up the background workers in the separate nodes. not sure the right way to handle that

gateway/uplink.go
```diff
@@ -98,7 +99,31 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 		device.Downlinks = device.Downlinks[1:]
 	}
 
-	if downlinkPayload != nil || len(requests) > 0 || sendAck {
+	// set the duty cycle in first downlink if in EU region.
+	setDutyCycle := false
+	if device.Region == regions.EU && device.FCntDown == 0 {
```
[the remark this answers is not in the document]
oliviamiller: they should always match so we could use either, I feel like its better to use the region of the device we're receiving the uplink from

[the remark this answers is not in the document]
JohnN193: yeah i guess thats kinda my point, this field isn't from the node. the `device` here is just from the gateway struct, which should always match the gateway. 

when we eventually configure multiple gateways this is could be even more of an issue, as the two gateways could have different regions configured for whatever reason.

randhid (commented):
Have we verified the following through manual testing:
1. No hang on viam server after a reconfigure
2. No hang on viam server after a module restart.
3. Successful reconfiguration of any of our nodes.
4. Successful reconfiguration with data capture enabled - this shows that the code is functional even with a concurrent call to Readings as a reconfigure is called.
5. Successful restart with data capture enabled.

gateway/gateway.go
```diff
@@ -172,13 +172,11 @@ type gateway struct {
...
 
-	logReader *os.File
-	logWriter *os.File
-	dataFile  *os.File
-	dataMu    sync.Mutex
-
-	region     regions.Region
+	logReader  *os.File
+	logWriter  *os.File
+	db         *sql.DB // store device information/keys for use across restarts in a database
```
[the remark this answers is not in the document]
oliviamiller: im not sure why the dataFile being removed is being shown as a diff, it is already removed in main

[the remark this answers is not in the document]
JohnN193: it is no longer shown as a diff on my end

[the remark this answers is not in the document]
randhid: Ah - the culprit may infact be moving `region` under `regionInfo` and git is showing the lines around that as a diff. It sometimes does that - resolving.

draginolht65n/sensor.go
JohnN193: so it would be gross, but within `ReconfigureWithConfig` you could technically unlock the mutex to run `GetAndUpdateDeviceInfo(ctx)`, then lock again before starting the background thread. Not sure if theres another way to solve it with our setup

oliviamiller: Yeah i considered doing that but thought it was gross, but fine with changing it to that if we want to remove the repeated code

JohnN193 (approved):
LGTM once the thread start is resolved.

draginolht65n/sensor.go
oliviamiller: Spoke with john offline: ended up putting the start of the thread in reconfigureWithConfig as we can start it without holding the lock, and keeping the GetAndUpdateDeviceInfo in the node's reconfigure function as it needs the lock.

oliviamiller:
> Have we verified the following through manual testing:
> 
> 1. No hang on viam server after a reconfigure
> 2. No hang on viam server after a module restart.
> 3. Successful reconfiguration of any of our nodes.
> 4. Successful reconfiguration with data capture enabled - this shows that the code is functional even with a concurrent call to Readings as a reconfigure is called.
> 5. Successful restart with data capture enabled.

confirmed all these cases work as expected

oliviamiller:
> > > Have we verified the following through manual testing:
> > > 
> > > 1. No hang on viam server after a reconfigure
> > > 2. No hang on viam server after a module restart.
> > > 3. Successful reconfiguration of any of our nodes.
> > > 4. Successful reconfiguration with data capture enabled - this shows that the code is functional even with a concurrent call to Readings as a reconfigure is called.
> > > 5. Successful restart with data capture enabled.
> > 
> > 
> > confirmed all these cases work as expected
> 
> Can you describe the result for each test? It is good to document things explicitly in general.

1. Reconfiguring gateway restarts hardware and gateway continues parsing uplinks.
2. all module components shut down and restart successfully after restarting the module. 
3. reconfiguring the uplink_interval_min on a node component triggers a interval downlink to be added to the queue, and errors if its below the minimum interval (won't be able to see the error in logs until next rdk release due to [bug](https://viam.atlassian.net/browse/RSDK-10368))
4. tests 1 and 3 behave the same way when data capture is enabled on gateway or node component. 
5. viam-server and module is shutting down and restarting cleanly with data capture enabled.

randhid (approved):

PR #52, opened 2025-04-02
move gateway to only use db for node information

JohnN193:


gateway/sqlite.go
```diff
@@ -50,8 +146,15 @@ func (g *gateway) insertOrUpdateDeviceInDB(ctx context.Context, device deviceInf
...
-		`devices(devEui, appSKey, nwkSKey, devAddr, fCntDown, nodeName, minUplinkInterval) VALUES(?, ?, ?, ?, ?, ?, ?);`
+	cmd := "insert or replace into devices(" +
+		devEUIDBKey + ", " +
+		appSDBKey + ", " +
+		nwkSDBKey + ", " +
+		devAddrDBKey + ", " +
+		fCntDownDBKey + ", " +
+		nodeNameDBKey + ", " +
+		minUplinkIntervalDBKey +
+		") VALUES(?, ?, ?, ?, ?, ?, ?);"
```
JohnN193, line 157: it looks like this because fmt.Sprintf is not safe with sqlite. so thats fun

gateway/gateway.go
```diff
@@ -557,104 +534,88 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+		}
 	}
 
 	// Update the device in the map.
 	g.devices[device.NodeName] = device
 	return nil
 }
 
-// mergeNodes merge the fields from the oldNode and the newNode sent from reconfigure.
-func mergeNodes(newNode, oldNode *node.Node) (*node.Node, error) {
```
JohnN193, line 602: since moving to the db file this function is now redundant

gateway/gateway.go
```diff
@@ -166,7 +154,7 @@ type gateway struct {
 	lastReadings map[string]interface{} // map of devices to readings
 	readingsMu   sync.Mutex
 
-	devices map[string]*node.Node // map of node name to node struct
+	devices map[string]gatewayNode // map of node name to node struct
```
JohnN193, line 157: the initial change was to simplify how `devices` was being used and to make it safe. now that we want to move to sqlite this change isn't super relevant anymore, but I still wanted to decouple the node struct from the gateway logic as it was causing some confusion

gateway/sqlite.go
```diff
@@ -26,20 +26,85 @@ var (
 	errDBClosed         = errors.New("error gateway is closed")
 )
 
+// Node defines a lorawan node device.
+type gatewayNode struct {
```
nicksanford, line 30: I don't understand the distinction between a deviceInfo and a gatewayNode

Is a `gatewayNode` stored in the DB? If not, why is it in the `gateway/sqlite.go` file?

gateway/sqlite.go
```diff
@@ -26,20 +26,85 @@ var (
...
+	nwkSDBKey              = "nwkSKey"
+	devAddrDBKey           = "devAddr"
+	fCntDownDBKey          = "fCntDown"
+	nodeNameDBKey          = "nodeName"
+	minUplinkIntervalDBKey = "minUplinkInterval"
+	appKeyDBKey            = "appKey"
+	decoderPathDBKey       = "decoderPath"
+	joinTypeDBKey          = "joinType"
+	fPortDBKey             = "fPort"
+	isRegisteredDBKey      = "isRegistered"
```
nicksanford, line 60: where are these fields used?

gateway/sqlite.go
```diff
@@ -22,24 +23,119 @@ var (
...
+						continue
+					}
+					return err
+				}
+			}
+		}
+	}
+
+	// unregister devices
+	cmd := "UPDATE devices SET " + isRegisteredDBKey + "=0"
```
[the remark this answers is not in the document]
JohnN193: currently it does not. no other code references isRegistered.

in a future pr setting this to 0 will prep the machine to "register" sensors that we want to connect. 

in other terms it allows us to differentiate between currently configured sensors and sensors that were configured in the past. the do command `register_device` will set these flags to 1

gateway/uplink.go
```diff
@@ -103,10 +102,15 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 			g.logger.Warnf("Sensor %v must be reset to support new features. "+
 				"Please physically restart the sensor to enable downlinks", device.NodeName)
 		} else {
-			payload, err := g.createDownlink(ctx, device, downlinkPayload, requests, sendAck, snr, sf)
+			payload, err := g.createDownlink(ctx, &device, downlinkPayload, requests, sendAck, snr, sf)
 			if err != nil {
 				return "", map[string]interface{}{}, fmt.Errorf("failed to create downlink: %w", err)
 			}
+
+			g.logger.Info("yo devices: ", g.devices)
```
nicksanford, line 110: should this be removed?

nicksanford (approved):

gateway/gateway.go
```diff
@@ -481,32 +469,20 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
-				g.devices[node.NodeName] = mergedNode
-			}
 			// Check if the device is in the persistent data file, if it is add the OTAA info.
-			deviceInfo, err := g.findDeviceInDB(ctx, hex.EncodeToString(g.devices[node.NodeName].DevEui))
+			deviceInfo, err := g.findDeviceInDB(ctx, node.DevEui)
 			if err != nil {
 				if !errors.Is(err, errNoDeviceInDB) {
 					return nil, fmt.Errorf("error while searching for device in file: %w", err)
 				}
+				g.devices[node.NodeName] = node
```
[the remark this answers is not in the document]
JohnN193: we should no longer need to set this on error. when I started this pr the "next step" would only update if the device was not found in the database, and this line was intended to handle if it was not found. 

now we always update the devices in the next function so we should no longer need this

gateway/sqlite.go
```diff
@@ -26,20 +26,85 @@ var (
...
 	filePathDB := filepath.Join(pathPrefix, "devicedata.db")
 	db, err := sql.Open("sqlite3", filePathDB)
 	if err != nil {
 		return err
 	}
-	// create the table if it does not exist
-	// if we want to change the fields in the table, a migration function needs to be created
-	cmd := `create table if not exists devices(devEui TEXT NOT NULL PRIMARY KEY,` +
-		` appSKey TEXT, nwkSKey TEXT, devAddr TEXT, fCntDown INTEGER, nodeName TEXT, minUplinkInterval REAL);`
+
```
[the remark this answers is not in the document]
JohnN193: thanks! will add a comment for this

gateway/sqlite.go
```diff
@@ -26,20 +26,85 @@ var (
...
+		} else {
+			cmd += ","
+		}
+	}
 	if _, err = db.ExecContext(ctx, cmd); err != nil {
 		return err
 	}
+
+	// unregister devices
+	cmdUnregister := "UPDATE devices SET " + isRegisteredDBKey + "=0"
```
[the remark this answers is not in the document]
JohnN193: based on tests yes

gateway/sqlite.go
```diff
@@ -26,20 +26,85 @@ var (
...
 	db, err := sql.Open("sqlite3", filePathDB)
 	if err != nil {
 		return err
 	}
-	// create the table if it does not exist
-	// if we want to change the fields in the table, a migration function needs to be created
-	cmd := `create table if not exists devices(devEui TEXT NOT NULL PRIMARY KEY,` +
-		` appSKey TEXT, nwkSKey TEXT, devAddr TEXT, fCntDown INTEGER, nodeName TEXT, minUplinkInterval REAL);`
+
+	cmd := `create table if not exists devices(`
```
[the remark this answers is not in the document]
JohnN193: 1. succeeds - covered in tests
2. succeeds but other parts of the code will fail - not covered in tests
3. succeeds  - covered in tests

gateway/sqlite.go
JohnN193: gatewayNode's information will all be stored in the db at some point. as we progress with the migration ideally this struct will be removed, and I placed it in sqlite.go so it would be easier to track what fields it should own.

gateway/sqlite.go
JohnN193: these fields are placeholders, meant to be used instead of `gatewayNode`. equivalent code is currently used in `join.go`, `uplink.go` and `downlink.go`

gateway/sqlite.go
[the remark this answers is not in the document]
JohnN193: https://github.com/viam-modules/lorawan/pull/52#discussion_r2027513876

randhid (approved):

gateway/gateway.go
```diff
@@ -557,104 +533,88 @@ func (g *gateway) DoCommand(ctx context.Context, cmd map[string]interface{}) (ma
...
+			// if join type is ABP get the new appSKey and addr from the new config.
+		default:
+			// we should already have validated the JoinType, but still throw an error here if this hits.
+			return errUnexpectedJoinType
+		}
 
-	device.AppSKey = appsKey
-	device.Addr = savedAddr
-	device.NwkSKey = nwksKey
+		device.NwkSKey = d.NwkSKey
```
oliviamiller, line 561: this should go in the joinTypeOTAA i think? since the nwkSKey will be provided by the user in the config for ABP

gateway/uplink.go
```diff
@@ -103,10 +102,14 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
 				"Please physically restart the sensor to enable downlinks", device.NodeName)
 		} else {
-			payload, err := g.createDownlink(ctx, device, downlinkPayload, requests, sendAck, snr, sf)
+			payload, err := g.createDownlink(ctx, &device, downlinkPayload, requests, sendAck, snr, sf)
 			if err != nil {
 				return "", map[string]interface{}{}, fmt.Errorf("failed to create downlink: %w", err)
 			}
+
+			// update state in gatewayNodes
+			g.devices[device.NodeName] = device
```
oliviamiller, line 111: kind of confused what this is doing, is the` device ` getting modified in createDownlink?

JohnN193: yes, `device` gets an updated `FCntDown` within createDownlink, which I need to reflect in g.devices now that we are no longer using a pointer to that part of the map

gateway/gateway.go
JohnN193: ahh gotcha, I remembered that was a thing but wasn't sure.

gateway/uplink.go
oliviamiller: could we move this line to createDownlink to make it more clear?

oliviamiller (approved):

JohnN193:
waiting to merge this until after [multiple gateways](https://viam.atlassian.net/browse/RSDK-10093) is completed to avoid merge conflicts & to get the next lorawan release out with the sol eng asks(eu regulation support & multiple gateways).

PR #53, opened 2025-04-11
RSDK-10388 Add support for dragino water quality sensor 

oliviamiller:
Model added for dragino WQS-LB. Added dragino package to share code for common downlink messages. Added wqs calibration downlinks, which a user can send to the device using the UI. 

Tested that the dragino lht65n still works and the wqs builds successfully. Don't have the hardware yet to fully test WQS.


dragino/downlink.go
```diff
@@ -0,0 +1,36 @@
...
+// SendIntervalDownlink sends an interval update command to a Dragino sensor.
+func SendIntervalDownlink(ctx context.Context, n *node.Node, intervalMin float64, testOnly bool) (map[string]interface{}, error) {
+	return n.SendIntervalDownlink(ctx, node.IntervalRequest{
+		IntervalMin:  intervalMin,
+		PayloadUnits: node.Seconds,
+		Header:       IntervalHeader,
+		NumBytes:     IntervalBytes,
+		TestOnly:     testOnly,
+	})
+}
```
JohnN193, line 36: instead of making these functions, I think we should just have functions called `CreateIntervalDownlinkReq `that return the `node.IntervalRequest` struct. then still call n.node.SendIntervalDownlink(CreateIntervalDownlinkReq()) in the sensors. That way we don't have to pass around the node pointer.

since the ResetRequest is always the same we can prob just make that into a constant rather than a function.

draginowqslb/sensor.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/viam-modules/gateway/dragino"
+	"github.com/viam-modules/gateway/node"
+	"go.viam.com/rdk/components/sensor"
+	"go.viam.com/rdk/logging"
+	"go.viam.com/rdk/resource"
+)
+
+const (
+	decoderURL = "https://raw.githubusercontent.com/dragino/dragino-end-node-decoder/" +
+		"5a2855dbddba7977e06ba710f33fbee27de124e5/WQS-LB/WQS-LB_ChirpstackV4_Decoder.txt"
```
JohnN193, line 18: nice we found this one! sucks we can't find the decoder for the other dragino

README.md
```diff
@@ -361,6 +362,80 @@ This command will send a generic downlink payload to the gateway. the string is
 }
 ```
 
+## Calibrate your `dragino-WQS-LB`
```
JohnN193, line 365: this will not appear on the markdown in app. is that intended?

README.md
```diff
@@ -361,6 +362,80 @@ This command will send a generic downlink payload to the gateway. the string is
...
+The WQS-LB water quality sensor should be calibrated upon first use. The calibration can be completed using do commands on the UI.
```
JohnN193, line 366: add a link to the guide you followed

JohnN193 (approved):
a few small things but otherwise lgtm

draginolht65n/sensor.go
```diff
@@ -129,8 +130,7 @@ func (n *LHT65N) Reconfigure(ctx context.Context, deps resource.Dependencies, co
 	// set the interval if one was provided
 	// we do not send a default in case the user has already set an interval they prefer
 	if cfg.Interval != nil && *cfg.Interval != 0 {
-		req := node.IntervalRequest{IntervalMin: *nodeCfg.Interval, PayloadUnits: node.Seconds, Header: "01", NumBytes: 3, TestOnly: false}
-		if _, err := n.node.SendIntervalDownlink(ctx, req); err != nil {
+		if _, err := dragino.SendIntervalDownlink(ctx, &n.node, *nodeCfg.Interval, false); err != nil {
```
randhid, line 133: nice cleanup.

draginowqslb/sensor.go
randhid: Is this helpful?
https://www.dropbox.com/scl/fo/ztlw35a9xbkomu71u31im/AIeuB8ots6IE61ihETFhC3Y/LoRaWAN%20End%20Node/LHT65N_LHT65S/Decoder?dl=0&preview=LHT65N%20Chirpstack%20%204.0%20decoder.txt&rlkey=ojjcsw927eaow01dgooldq3nu&subfolder_nav_tracking=1

I don't know if their unholy idea of hosting them all on dropbox works as well as a gist.

draginowqslb/sensor.go
```diff
@@ -0,0 +1,243 @@
...
+			return map[string]interface{}{}, fmt.Errorf("expected float64 got %v", reflect.TypeOf(ph))
+		}
+		payload := "FB"
+		switch phFloat {
+		case 9:
+			payload += "09"
+		case 6:
+			payload += "06"
+		case 4:
+			payload += "04"
```
randhid, line 172: Can you comment over these cases what these mean for the ph sensor - I'm assuming these are 1:1 to the acid/neutral/base values?

draginowqslb/sensor.go
```diff
@@ -0,0 +1,243 @@
...
+	if ec, ok := cmd["calibrate_ec"]; ok {
+		ecFloat, ok := ec.(float64)
+		if !ok {
+			return map[string]interface{}{}, fmt.Errorf("expected float64 got %v", reflect.TypeOf(ec))
+		}
+		payload := "FD"
+		switch ecFloat {
+		case 1:
+			payload += "01"
+		case 10:
```
randhid, line 188: Similarly here and for the rest of the calibration DoCommands - comment which step corresponds to which solution state.

We can be lazy this way and not look at the reamde

draginowqslb/sensor.go
```diff
@@ -0,0 +1,243 @@
...
+			payload += "02"
+		case 400:
+			payload += "04"
+		case 600:
+			payload += "06"
+		case 800:
+			payload += "08"
+		case 1000:
+			payload += "0A"
+		default:
```
randhid, line 217:
```suggestion
		// each case here corresponds to an NTU value, corresponding to each step in the calibration instructions of the READMe 
		case 0:
			payload += "00"
		case 200:
			payload += "02"
		case 400:
			payload += "04"
		case 600:
			payload += "06"
		case 800:
			payload += "08"
		case 1000:
			payload += "0A"
		default:
```

randhid (commented):
One suggestion about adding bools to inform the user of the state of the sensor calibration- which I think would be good eventual behaviour from this driver.

If it is difficult to tell exactly what the best user experience is without physical devices on hand, I'm happy to defer it to a follow-up ticket. I expect a user will ask to know whether the probe is calibrated in the future anyway. 

A question - are there other downlinks to send to these sensors that tell you whether the probe is out of calibration at all?

README.md
oliviamiller: I thought back when markdown docs was introduced eliot requested that only config attributes be in it so I didn't include it. But im fine with including it if we prefer

draginowqslb/sensor.go
```diff
@@ -0,0 +1,243 @@
...
+		return nil, err
+	}
+	return deps, nil
+}
+
+// WQSLB defines the WQS-LB sensor implementation.
+type WQSLB struct {
+	resource.Named
+	logger logging.Logger
+	node   node.Node
```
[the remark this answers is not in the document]
JohnN193: I could be misreading the manual, but there does not appear to be an [uplink message](https://wiki.dragino.com/xwiki/bin/view/Main/User%20Manual%20for%20LoRaWAN%20End%20Nodes/WQS-LB--LoRaWAN_Water_Quality_Sensor_Transmitter_User_Manual/#H2.2Uplinkpayload) that returns whether the sensor is calibrated.

README.md
JohnN193: DoCommands are also listed in the markdown, and since some probes won't function correctly without calibration I think we should make it easy for users to find that information.

draginowqslb/sensor.go
JohnN193: I don't remember which decoder we use currently. can test this one quick next week

draginowqslb/sensor.go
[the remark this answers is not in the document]
randhid: Then I guess we have no way to tell if a sensor is calibrated.

If we follow my design - we can only tell if a sensor is calibrated within the lifetime of the resource or viam-server.

Adding this state across restarts is a little complicated - we'd need to figure out where to store it and that's up in the air.

Thanks for looking at the manual John. I wish it had an uplink that would solve this for us - but I guess nothing to do right now.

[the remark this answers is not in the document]
oliviamiller: Added bools and warnings to the readings, gets tricky though because the sensor won't actually be calibrated until the downlink gets sent after the next uplink. Also will get messed up upon restart of the the module, didn't want to save these in sqlite since they are specific to this sensor but don't know where to store them.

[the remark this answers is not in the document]
randhid: I'm so sorry - when i said
> but I guess nothing to do right now.
I did mean that we should take no action.

Thank you for being proactive. IF you really want to add this information and you think this is a good idea I would do the following:

NOTE: It is toally okay to go bak to your previous design and not do this. I want to be clear about that. This is just a suggestion

1. group together the bools into structs for each calibration
So for example:
```go
 type phCalState struct {       
        step1Complete  bool
	step2Complete  bool
	step3Complete  bool
}
```

2.  use helper functions on the structs themselves to return true or false rather than checking each one alone
```go
 (pcs *phCalState) checkClaibrated() bool {
      return pcs.step1Complete && Step2Complete && Step3Complete
 }
```

3. Just check each calibration helper at the top of The DoCommand and warn if any return false.

draginowqslb/sensor.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// turbidity calibration
+	if t, ok := cmd["calibrate_t"]; ok {
+		tFloat, ok := t.(float64)
+		if !ok {
+			return map[string]interface{}{}, fmt.Errorf("expected float64 got %v", reflect.TypeOf(t))
+		}
+		payload := "FE"
+		// each case here corresponds to an NTU value,
+		// corresponding to each step in the calibration instructions of the READMe
```
randhid, line 219:
```suggestion
		payload := "FE"
		// each case here corresponds to an NTU value,
		// corresponding to each step in the calibration instructions of the README
```

meta.json
```diff
@@ -26,9 +26,15 @@
...
         "model": "viam:lorawan:dragino-LHT65N",
-        "markdown_link": "README.md#configure-your-viamlorawandragino-lht65n",
+        "markdown_link": "README.md#configure-your-dragino-sensor",
         "short_description": "A sensor component for a dragino LHT65N LoRaWAN node."
      },
+     {
+      "api": "rdk:component:sensor",
+      "model": "viam:lorawan:dragino-WQS-LB",
+      "markdown_link": "README.md#configure-your-dragino-sensor",
+      "short_description": "A sensor component for a dragino WQS-LB LoRaWAN node."
```
randhid, line 36: @JohnN193 where in the meta.json do we add the DoCommands from the readme?

randhid (approved):
If we could try to find where we can add the Docommands in the README that would be great.

Let's create an arc for Felix to get feedback before we promote this to a full release of the Lorawan module.

meta.json
JohnN193: It’s part of the markdown link. The module generator includes DoCommands in the readme.

JohnN193: This module already displays most DoCommands via the markdown link. This pr was just not including them yet

randhid: Thanks for explaining!

PR #55, opened 2025-04-28
RSDK-10585 Add USB support to generic gateway

oliviamiller:
Adds optional attribute `path` to generic gateway that can be used to specify the serial or SPI path. the com type is determined based on the value of `path` attribute. `spi_bus` can also still be used for SPI to avoid breaking anything.

Tested on the rak USB concentrators and confirmed SPI still works with waveshare hat.

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -39,6 +39,7 @@ const (
 var (
 	// Config validation errors.
 	errInvalidSpiBus = errors.New("spi bus can be 0 or 1 - default 0")
+	errSPIAndUSB     = errors.New("cannot have both spi_bus and path attributes, add path for USB or spi_bus for SPI gateways")
 
 	// Gateway operation errors.
 	errUnexpectedJoinType = errors.New("unexpected join type when adding node to gateway")
@@ -138,6 +139,10 @@ func (conf *Config) Validate(path string) ([]string, error) {
 
 	if conf.Bus != nil && *conf.Bus != 0 && *conf.Bus != 1 {
 		return nil, resource.NewConfigValidationError(path, errInvalidSpiBus)
+	}
+
+	if conf.Bus != nil && conf.Path != "" {
+		return nil, resource.NewConfigValidationError(path, errSPIAndUSB)
 	}
 
 	if len(conf.BoardName) == 0 {
@@ -312,15 +317,21 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		comType = lorahw.SPI
 	}
 
-	// user provided an SPI or serial device path
+	// user provided a serial device path
 	if cfg.Path != "" {
 		path = cfg.Path
-		if strings.Contains(path, "spi") {
-			comType = lorahw.SPI
-		} else {
-			comType = lorahw.USB
-		}
+		comType = lorahw.USB
+
 	}
+	var comTypeString string
+	switch comType {
+	case lorahw.SPI:
+		comTypeString = "SPI"
+	case lorahw.USB:
+		comTypeString = "USB"
+	}
+
+	g.logger.Infof("starting %s gateway connected to path %s", comTypeString, path)
 
 	if err := lorahw.SetupGateway(comType, path, region); err != nil {
 		return fmt.Errorf("failed to set up the gateway: %w", err)
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -88,7 +88,7 @@ var noReadings = map[string]interface{}{"": "no readings available yet"}
 
 // Config describes the configuration of the gateway.
 type Config struct {
-	Bus       int    `json:"spi_bus,omitempty"`
+	Bus       *int   `json:"spi_bus,omitempty"`
 	PowerPin  *int   `json:"power_en_pin,omitempty"`
 	ResetPin  *int   `json:"reset_pin"`
 	BoardName string `json:"board"`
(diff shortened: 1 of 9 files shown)


gateway/gateway.go
```diff
@@ -295,7 +297,32 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+
+	// user provided an SPI or serial device path
+	if cfg.Path != "" {
+		path = cfg.Path
+		if strings.Contains(path, "spi") {
+			comType = lorahw.SPI
+		} else {
+			comType = lorahw.USB
+		}
+	}
```
JohnN193, line 323: are we expecting user's machines to have a different spi path? 

my worry is users set a Bus and USB path for some reason, which would put us in a bad state, and that the path should only be used for usb.

JohnN193 (commented):
Main concern is whether `path` should also support the spi path. looks good outside of that

gateway/gateway.go
JohnN193: if we aren't worried about people misconfiguring their device then its probably fine to leave as is, just add a debug log that calls out what connection method the code will use

gateway/gateway.go
```diff
@@ -134,7 +135,8 @@ func (conf *Config) Validate(path string) ([]string, error) {
 	if conf.ResetPin == nil {
 		return nil, resource.NewConfigValidationFieldRequiredError(path, "reset_pin")
 	}
-	if conf.Bus != 0 && conf.Bus != 1 {
+
+	if conf.Bus != nil && *conf.Bus != 0 && *conf.Bus != 1 {
```
randhid, line 139: Can you also add a check in Validate to not let people configure both USB and SPI together?

I don't know if the 16-channel gateway can mix and match, but I think the normal hats won't be able to.

gateway/gateway.go
```diff
@@ -88,11 +88,12 @@ var noReadings = map[string]interface{}{"": "no readings available yet"}
 
 // Config describes the configuration of the gateway.
 type Config struct {
-	Bus       int    `json:"spi_bus,omitempty"`
+	Bus       *int   `json:"spi_bus,omitempty"`
 	PowerPin  *int   `json:"power_en_pin,omitempty"`
 	ResetPin  *int   `json:"reset_pin"`
 	BoardName string `json:"board"`
 	Region    string `json:"region_code,omitempty"`
+	Path      string `json:"path,omitempty"`
```
randhid, line 96: Can we try to find this automatically using the /serial/by-id/ or /serial/by-path/ routes?

gateway/gateway.go
```diff
@@ -295,7 +297,32 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
 
-	if err := lorahw.SetupGateway(cfg.Bus, region); err != nil {
+	var path string
+	var comType lorahw.ComType
+
+	// no bus or path provided, use default spi bus 0
+	if cfg.Bus == nil && cfg.Path == "" {
+		comType = lorahw.SPI
+		path = "/dev/spidev0.0"
+	}
```
randhid, line 307: log this.

gateway/gateway.go
```diff
@@ -295,7 +297,32 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+
+	// user provided a spi bus
+	if cfg.Bus != nil {
+		path = fmt.Sprintf("/dev/spidev0.%d", *cfg.Bus)
+		comType = lorahw.SPI
+	}
```
randhid, line 313: kinda nice.

randhid (commented):
Q: I like that we've expanded generic to encompass all things gateway - but we should be careful about not stuffing too much into it.

Do you want to make a new USB gateway model, or keep these features in generic?

gateway/gateway.go
oliviamiller: Yeah i thought it was convenient to use path for both but can be confusing so I can make path be just for usb

gateway/gateway.go
oliviamiller: we could but if there are multiple serial connections (like on the rak) then we wouldn't know which one to use

oliviamiller:
> Q: I like that we've expanded generic to encompass all things gateway - but we should be careful about not stuffing too much into it.
> 
> Do you want to make a new USB gateway model, or keep these features in generic?

i'd rather have just one generic model but don't feel super strongly. Fine with splitting to two models if we prefer

JohnN193:
> > Q: I like that we've expanded generic to encompass all things gateway - but we should be careful about not stuffing too much into it.
> > Do you want to make a new USB gateway model, or keep these features in generic?
> 
> i'd rather have just one generic model but don't feel super strongly. Fine with splitting to two models if we prefer

are we planning on adding a rak model? or are we just going to let them configure with the generic gateway until there's a need for 16 channel?

oliviamiller:
> > > Q: I like that we've expanded generic to encompass all things gateway - but we should be careful about not stuffing too much into it.
> > > Do you want to make a new USB gateway model, or keep these features in generic?
> > 
> > 
> > i'd rather have just one generic model but don't feel super strongly. Fine with splitting to two models if we prefer
> 
> are we planning on adding a rak model? or are we just going to let them configure with the generic gateway until there's a need for 16 channel?

Yes still planing on adding the rak model

JohnN193 (approved):

gateway/gateway.go
randhid: Okay, it is a configuration pain poitn as you have to ssh in, but perhaps we could offer a discovery service.

randhid (approved):

PR #56, opened 2025-05-16
RSDK-10724 validate that serial path is valid

oliviamiller:
Validate that the serial path exists during validation. Tested manually on rak


JohnN193 (approved):
LGTM, guessing we wanted to do the simlinks/replacement stuff later?

PR #58, opened 2025-05-16
RSDK-10686 using serial by-path causes buffer overflow error 

oliviamiller:
converting the by-path or by-id path to the symlink so thats its small enough to fit into the c buffer. Tested manually on the rak

### DIFF
--- a/node/node_test.go
+++ b/node/node_test.go
@@ -271,6 +271,8 @@ func TestNewNode(t *testing.T) {
 	test.That(t, n, test.ShouldNotBeNil)
 
 	testNode = n.(*Node)
+	// lock to prevent a data race w the pollGateway routine
+	testNode.reconfigureMu.Lock()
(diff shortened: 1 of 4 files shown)


node/node_test.go
```diff
@@ -250,6 +250,8 @@ func TestNewNode(t *testing.T) {
 	test.That(t, testNode.NodeName, test.ShouldEqual, testNodeName)
 	test.That(t, testNode.JoinType, test.ShouldEqual, JoinTypeOTAA)
 	test.That(t, testNode.DecoderPath, test.ShouldEqual, testDecoderPath)
+	err = n.Close(ctx)
```
oliviamiller, line 253: This (seemingly not relevant to this code change) test kept failing in CI so fixed to pass checks

JohnN193 (approved):

PR #60, opened 2025-05-22
RSDK-10773 filter out invalid data from WQS-LB probes

oliviamiller:
min and max values for each key are based on the ranges from the user manual. The reading will be removed from captured data, and other calls will return an "INVALID" string. Tested manually that the EC K10 invalid data returns string/captured data does not contain the EC K10 value.

### DIFF
--- a/draginowqslb/sensor.go
+++ b/draginowqslb/sensor.go
@@ -17,6 +17,10 @@ import (
 const (
 	decoderURL = "https://raw.githubusercontent.com/dragino/dragino-end-node-decoder/" +
 		"5a2855dbddba7977e06ba710f33fbee27de124e5/WQS-LB/WQS-LB_ChirpstackV4_Decoder.txt"
+
+	// Ranges for each probe were found in the user manual:
+	// https://wiki.dragino.com/xwiki/bin/view/Main/User%20Manual%20for%20LoRaWAN%20End%20Nodes
+	// /WQS-LB--LoRaWAN_Water_Quality_Sensor_Transmitter_User_Manual/
 	ecK10Key     = "EC_K10"
 	ecK10Min     = 20.
 	ecK10Max     = 20000.
@@ -36,6 +40,15 @@ const (
 	turbidityMin = 0.1
 	turbidityMax = 10000.
 )
+
+var ProbeRanges = []probeRange{
+	{ecK10Key, ecK10Min, ecK10Max},
+	{ecK1Key, ecK1Min, ecK1Max},
+	{phKey, phMin, phMax},
+	{orpKey, orpMin, orpMax},
+	{doKey, doMin, doMax},
+	{turbidityKey, turbidityMin, turbidityMax},
+}
 
 var (
 	// Model represents the WQS-LB sensor model.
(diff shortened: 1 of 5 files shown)


draginowqslb/sensor.go
```diff
@@ -140,16 +159,47 @@ func (n *WQSLB) Close(ctx context.Context) error {
...
+		reading = sanitizeReading(reading, extra, probeRange)
+	}
+
 	return reading, nil
 }
 
+func sanitizeReading(reading, extra map[string]interface{}, limits probeRange) map[string]interface{} {
+	if val, ok := reading[limits.key].(float64); ok && (val > limits.max || val < limits.min) {
+		// remove reading if from data capture
+		if extra[data.FromDMString] == true {
```
JohnN193, line 194: is this okay if extra gets passed in as nil? i think rdk prevents this from happening but just wanna check

draginowqslb/sensor.go
```diff
@@ -140,16 +159,47 @@ func (n *WQSLB) Close(ctx context.Context) error {
...
 	}
 
+	ranges := []probeRange{
+		{ecK10Key, ecK10Min, ecK10Max},
+		{ecK1Key, ecK1Min, ecK1Max},
+		{phKey, phMin, phMax},
+		{orpKey, orpMin, orpMax},
+		{doKey, doMin, doMax},
+		{turbidityKey, turbidityMin, turbidityMax},
+	}
```
JohnN193, line 182: should this live on the struct or live as a global? seems odd to define it here

draginowqslb/sensor.go
```diff
@@ -140,16 +159,47 @@ func (n *WQSLB) Close(ctx context.Context) error {
...
+
+	for _, probeRange := range ranges {
+		reading = sanitizeReading(reading, extra, probeRange)
+	}
+
 	return reading, nil
 }
 
+func sanitizeReading(reading, extra map[string]interface{}, limits probeRange) map[string]interface{} {
```
JohnN193, line 191: [q] should this move into node utils? 

I'm assuming other sensors will want a similar functionality but I don't know if we want the function to be the same. might be better to wait until we investigate those to move this

draginowqslb/sensor.go
```diff
@@ -9,13 +9,32 @@ import (
...
+	"go.viam.com/rdk/data"
 	"go.viam.com/rdk/logging"
 	"go.viam.com/rdk/resource"
 )
 
 const (
 	decoderURL = "https://raw.githubusercontent.com/dragino/dragino-end-node-decoder/" +
 		"5a2855dbddba7977e06ba710f33fbee27de124e5/WQS-LB/WQS-LB_ChirpstackV4_Decoder.txt"
+	ecK10Key     = "EC_K10"
+	ecK10Min     = 20.
```
JohnN193, line 21: add a link to the doc where we find these numbers

JohnN193 (approved):
LGTM. 

tried looking at the readme but i don't think theres a good spot to describe this behavior atm. unless we wanted to add another section after the DoCommand section that describes invalid data/other troubleshooting behaviors(the current troubleshooting section does not appear on configure cards)

draginowqslb/sensor.go
oliviamiller: we've been using this in gateway and generic node readings w no issue so i think its okay, we can add a nil check to be extra safe

draginowqslb/sensor.go
oliviamiller: I can make it a global

draginowqslb/sensor.go
oliviamiller: yeah we can move it when we add similar functionality to others

PR #62, opened 2025-05-23
RSDK-10357 Add RAK7391 model

oliviamiller:
Adds the rak7391 model that allows for configuration of two concentrators. Lots of repeated code with gateway package for now until we determine how we will support multiple gateways.


rak7391/rak7391.go
```diff
@@ -0,0 +1,996 @@
...
+	conn, err := grpc.NewClient(target, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return fmt.Errorf("error connecting to server: %w", err)
+	}
+
+	r.logger.Info("connected to client")
+
+	c.conn = conn
+	c.client = pb.NewSensorServiceClient(conn)
+	c.started = true
```
randhid, line 382: When is this ever set to false in this driver's code? Just on creation? So viam server restart?

randhid (commented):
I wish there wasn't so much repetition in the gateway packages, but I understand the code isn't known to be extensible at this point.

I would like some documentation of behaviour tests on this PR.

- What was the behaviour for reconfiguring?
- What was the behaviour across viam server restarts? Was valid data being captured?
- What was the behaviour at closure or module restart.
- What was the behaviour after adding and removing nodes?

- What was the behaviour after removing one of the concentrators from the config. Can each concentrator be removed, and do we observe the resource reconfiguring as expected?

JohnN193 (changes requested):
got abit lost in the review. short summary of thoughts:

regardless of how multiple gateways ends up happening, I think the gateway package should own handling multiple concentrators, rather than moving the logic to another package. 

most of the code you have written can be moved into the gateway package. the only thing that jumps out to me as "missing" if we were to move everything into the gateway package is logic to handle reconfigure, like what we did for the `waveshare_hat`

happy to talk more on this. i think "multiple concentrators"(or some idea similar to this) will probably keep coming up, so if we do want to wait until the multiple gateways refactor then thats fine with me. tbh multiple concentrators is actually kinda similar to a plan for multiple gateways that we had drawn up on the whiteboard

rak7391/rak7391.go
```diff
@@ -0,0 +1,996 @@
...
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create command struct: %w", err)
+		}
+
+		req := &v1.DoCommandRequest{
+			Command: cmdStruct,
+		}
+
+		_, err = r.concentrators[0].client.DoCommand(ctx, req)
```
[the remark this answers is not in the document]
oliviamiller: this was used to for debugging to see if the client is still connected, now that development is done I can just remove this, it prob won't need to be used again

oliviamiller:
> got abit lost in the review. short summary of thoughts:
> 
> regardless of how multiple gateways ends up happening, I think the gateway package should own handling multiple concentrators, rather than moving the logic to another package.
> 
> most of the code you have written can be moved into the gateway package. the only thing that jumps out to me as "missing" if we were to move everything into the gateway package is logic to handle reconfigure, like what we did for the `waveshare_hat`
> 
> happy to talk more on this. i think "multiple concentrators"(or some idea similar to this) will probably keep coming up, so if we do want to wait until the multiple gateways refactor then thats fine with me. tbh multiple concentrators is actually kinda similar to a plan for multiple gateways that we had drawn up on the whiteboard

Agree we should just move everything to gateway package and have gateway handle multiple concentrators, modifying this PR to do that

JohnN193 (commented):
Looks good, prob LGTM but gonna look again in the morning cuz my setup right now isn’t ideal

gateway/gateway.go
```diff
@@ -282,54 +323,85 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
-	if err != nil {
-		return err
+	if g.lastReadings == nil {
+		g.lastReadings = make(map[string]interface{})
 	}
 
+	g.concentrators = make([]*concentrator, 0)
+
+	switch conf.Model {
+	case ModelRak7391:
```
JohnN193, line 345: kinda wish there was another way for us to check if its a multi config outside of checking all of the model names here. not too sure what that would look like though

gateway/gateway.go
```diff
@@ -282,54 +323,85 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
...
+	switch region {
+	case regions.US, regions.Unspecified:
+		g.logger.Infof("configuring gateway for US915 band")
+		g.regionInfo = regions.RegionInfoUS
+		g.region = regions.US
+	case regions.EU:
+		g.logger.Infof("configuring gateway for EU868 band")
+		g.regionInfo = regions.RegionInfoEU
+		g.region = regions.EU
 	}
```
JohnN193, line 385: could this check live in the main Reconfigure function? gut feeling is no because the different configs

gateway/gateway.go
```diff
@@ -429,18 +527,21 @@ func (g *gateway) Reconfigure(ctx context.Context, deps resource.Dependencies, c
 		return fmt.Errorf("error connecting to server: %w", err)
 	}
 
-	g.logger.Info("connected to concentrator client")
-	g.conn = conn
-	g.concentratorClient = pb.NewSensorServiceClient(conn)
-	g.started = true
-	g.workers.Add(g.receivePackets)
+	g.logger.Info("connected to client")
```
JohnN193, line 530: lets switch this to debug and log the name of the concentrator

JohnN193 (approved):

randhid (approved):
Awesome job! Looks really really good!

PR #63, opened 2025-05-30
RSDK-10775 Support 16 channels 

oliviamiller:
support 16 channels on dual concentrators.

The module tells the nodes what channels to use by:
- the CFList in the join accept: if there are two concentrators we update the CFList to include 16 channels
- The linkADRReq downlink: This is sent if the node has already joined the network at server start, and at reconfigure if the number of concentrators change.

If only one concentrator is being used on the rak model, it will use channels 0-7. If both are configured, pcie2 will receive on 8-15.

16 channels is only supported in US frequency band.

Tested manually on rak and waveshare, confirmed build, reconfigure and close work as expected.


gateway/downlink.go
```diff
@@ -137,25 +139,30 @@ func (g *gateway) createDownlink(ctx context.Context,
 
 	payload = append(payload, devAddrLE...)
 
-	fopts := make([]byte, 0)
-
 	// Handle any mac commands sent in the uplink fopts.
 	if len(uplinkFopts) != 0 {
 		for _, b := range uplinkFopts {
+			// Break early if FOpts already has 15 bytes
```
JohnN193, line 145: how often can this case happen? if our fopts is too large which will we send?

gateway/gateway_test.go
```diff
@@ -728,9 +756,10 @@ func TestGetMultiConcentratorConfig(t *testing.T) {
 						ResetPin: &resetPin1,
 					},
 					{
-						Name:     "pcie2",
-						Bus:      &bus2,
-						ResetPin: &resetPin2,
+						Name:        "pcie2",
+						Bus:         &bus2,
+						ResetPin:    &resetPin2,
+						BaseChannel: 8,
```
JohnN193, line 762: would it be worth doing 0 & 1 instead of 0 & 8? then changing the name to ChannelGroup or something?

node/node.go
```diff
@@ -187,9 +187,10 @@ type Node struct {
 	JoinType      string
 	reconfigureMu sync.Mutex
 
-	FCntDown  uint16
-	FPort     byte     // for downlinks, only required when frame payload exists.
-	Downlinks [][]byte // list of downlink frame payloads to send
+	FCntDown    uint16
+	FPort       byte     // for downlinks, only required when frame payload exists.
+	Downlinks   [][]byte // list of downlink frame payloads to send
+	FoptsToSend [][]byte // list of mac commands to send
```
JohnN193, line 193: should we have a ticket for supporting mac commands from users?

gateway/downlink.go
oliviamiller: I wouldn't expect it to happen often (possibly never). If it does get over 15 bytes and we can't answer one of the requests the device should send the unanswered request again so we can answer it in a different downlink

gateway/gateway_test.go
oliviamiller: id rather keep it as channels rather than groups in case we want to make it configurable in the future. Then the user could use any of the 64 channels

JohnN193 (approved):

lorahw/gateway.c
```diff
@@ -49,14 +55,16 @@ int set_up_gateway(int com_type, char* path, int region) {
...
+    int base_freq;
     switch(region) {
         case 2:
             radio0_freq = EU_RADIO_0_FREQ;
             radio1_freq = EU_RADIO_1_FREQ;
             break;
         default:
-            radio0_freq = US_RADIO_0_FREQ;
-            radio1_freq = US_RADIO_1_FREQ;
+            base_freq = US_BASE_FREQ + (base_channel * 200000);
```
randhid, line 65: [q] Why does the base channel get added to the US base frequency but not the EU frequency?

gateway/uplink.go
```diff
@@ -109,21 +112,28 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
...
+		foptsToSend = device.FoptsToSend[0]
+		device.FoptsToSend = device.FoptsToSend[1:]
+	}
+
 	// set the duty cycle in first downlink if in EU region.
 	setDutyCycle := false
 	if g.region == regions.EU && device.FCntDown == 0 {
 		setDutyCycle = true
-		minInterval := calculateMinUplinkInterval(sf, len(phyPayload))
+		minInterval := calculateMinUplinkInterval(packet.DataRate, len(phyPayload))
```
randhid, line 125: nice cleanup.

gateway/uplink.go
```diff
@@ -109,21 +112,28 @@ func (g *gateway) parseDataUplink(ctx context.Context, phyPayload []byte, packet
 		device.Downlinks = device.Downlinks[1:]
 	}
 
+	var foptsToSend []byte
+	if len(device.FoptsToSend) > 0 {
+		foptsToSend = device.FoptsToSend[0]
```
randhid, line 117: add a comment on why we break out just the first byte - this is the byte that allows the sensor the receive a downlink?

randhid (approved):
A few nits and questions, but generally looks good to me.

Biggest question I have is why the base channel is only added to the US frequencies.

gateway/uplink.go
oliviamiller: its an array of bytes so we are getting the first byte array to send. We only send one mac command per downlink, will add a comment saying that.

PR #70, opened 2025-07-17
Add component name in low SNR warnings

hunter-volkman:
This PR moves the SNR validation from `receivePackets` to `parseDataUplink` after device identification, allowing us to include the component name in the log message:
```bash
7/16/2025, 5:06:49 PM warn rdk:component:sensor/gateway     gateway/uplink.go:63 packet from sensor-1 skipped due to low signal noise ratio: -13.25, min is -8.5   log_ts 2025-07-16T21:06:49.841Z
```

Currently, when packets are dropped due to low SNR (Signal-to-Noise Ratio), the log message doesn't identify which specific sensor/component is affected:
```bash
7/16/2025, 10:18:59 AM warn rdk:component:sensor/gateway     gateway/gateway.go:472 packet skipped due to low signal noise ratio: -11, min is -7.5   log_ts 2025-07-16T14:18:59.032Z
```

This makes it difficult to debug signal issues in deployments with multiple LoRaWAN sensors, as users can't tell which sensor needs attention.

After the change, the logging will show:

1. Known devices with low SNR:
```bash
packet from sensor-1 skipped due to low signal noise ratio: -13.25, min is -8.5   log_ts 2025-07-16T21:06:49.841Z
```

2. Unknown devices with low SNR (likely corrupted due to poor signal):
```bash
received packet from unknown device A84041EA01881AFF (likely due to poor SNR: -12, min is -7.5), ignoring
```

3. Unknown devices with good SNR (truly unknown devices):
```bash
received packet from unknown device B12345678901234, ignoring
```

### DIFF
--- a/gateway/gateway.go
+++ b/gateway/gateway.go
@@ -64,6 +64,7 @@ var (
 	errInvalidRegion              = errors.New("unrecognized region code, valid options are US915 and EU868")
 	errInvalidConcentratorsLength = errors.New("invalid concentrator length - should not happen")
 	errTimedOut                   = errors.New("timed out waiting for gateway to start")
+	errLowSNR = errors.New("packet SNR below minimum threshold")
 )
 
 // constants for MHDRs of different message types.
@@ -745,12 +746,6 @@ func (g *gateway) receivePackets(ctx context.Context) {
 					if isDuplicate {
 						continue
 					}
-					minSNR := sfToSNRMin[packet.DataRate]
-					if float64(packet.SNR) < minSNR {
-						g.logger.Warnf("packet skipped due to low signal noise ratio: %v, min is %v", packet.SNR, minSNR)
-						continue
-					}
-
 					g.handlePacket(ctx, packet, t, c)
 				}
 			}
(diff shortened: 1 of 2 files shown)


JohnN193 (approved):
LGTM

the issue we were worried about with snr was that low snrs would also have invalid dev addrs alot of the time. you will prob see more `errNoDevice` failures on gateways, but usually those devices are only off by like a number

gateway/uplink.go
```diff
@@ -57,6 +57,14 @@ func (g *gateway) parseDataUplink(ctx context.Context, packet lorahw.RxPacket, p
 		return "", map[string]interface{}{}, errNoDevice
```
nicksanford, line 64: @hunter-volkman would mind please:

1. Also checking whether `float64(packet.SNR) < minSNR`  is true in this block and if so change the debug to state that the SNR is bad. This will make it less ambiguous when the reason why this code doesn't know which device the packet came from is because the SNR is too poor that the packet is malformed. This will allow us to differentiate between a packet from a truly unknown device with good SNR (aka a device we don't care about) vs a device we potentially do care about, but has a poor signal.
2. Add the devAddrBE to both the `received packet from unknown device, ignoring` log and the log I requested in point 1.

gateway/gateway.go
```diff
@@ -745,12 +746,6 @@ func (g *gateway) receivePackets(ctx context.Context) {
 					if isDuplicate {
 						continue
 					}
-					minSNR := sfToSNRMin[packet.DataRate]
-					if float64(packet.SNR) < minSNR {
-						g.logger.Warnf("packet skipped due to low signal noise ratio: %v, min is %v", packet.SNR, minSNR)
```
JohnN193, line 750: so felix had initially asked us to make this a warning instead of an error. now that we are switching back to an error below we should prob let him know

JohnN193 (approved):

